PR #69, opened 2021-12-17
Card state test and static debugging

maevyn11:
Sorry for the chonky PR, this ties up most of my pre-demo workstreams. Mainly it adds the ability to do static debugging, including ways to call it manually and configure it in cmd, as well as outputing an apdu.log file that can be used in the Javacard debugger to inspect card errors. There is also some various cleanup and improvement of interfaces like PhononCard. Since there were so many changes going in at once I manually tested each command under cmd/ to ensure they all worked, which they do. Several were outdated or referred to unimplemented functionality, so I deleted those.


npmaile (approved):
LGTM

PR #71, opened 2022-01-12
added ability to transparently parse ECC compressed and uncompressed …

maevyn11:
…format


scripts/hexFormatter.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/python
```
npmaile, line 1: not sure this file needs to be committed to git.
If it is, I'd rather it be written in go so we don't have more languages than strictly necessary

maevyn11: Yea it was just simplest to write in python. We may need it again next time we update the keys, so I thought it'd be better to commit it now and not lose track of it. Don't think it really matters what language it's in as long as it's commonly available, since this is just a development utility and doesn't run along with the rest of the program. No different than the Makefile really.

PR #87, opened 2022-01-27
Separate connect pair

npmaile:
feel free to look at the code. 
it's not quite ready to go in yet.


remote/server.go
```diff
@@ -79,8 +80,8 @@ func handle(w http.ResponseWriter, r *http.Request) {
 	}
 	if !valid {
 		//TODO: use a real error, possibly from cert package
```
maevyn11, line 82: Still TODO or we happy with this error message now?

model/counterpartyCard.go
```diff
@@ -26,3 +27,41 @@ const (
 	StatusCardPair1Complete
 	StatusCardPair2Complete
 )
+
```
maevyn11, line 30: Think this message type should stay encapsulated in the remote package. Since we eventually want to implement multiple networking types and enable external apps to develop them I think we should keep the core client to card stuff as isolated as possible from the Jumpbox implementation.

repl/cardPairing.go
```diff
@@ -31,14 +31,14 @@ func cardPairLocal(c *ishell.Context) {
...
 	}
+
 	pairingCard := otherCards[selection]
 	c.Println("starting pairing with ", pairingCard.GetName())
 	remoteCard := orchestrator.NewLocalCounterParty(pairingCard)
-
-	err := activeCard.PairWithRemoteCard(remoteCard)
+	activeCard.RemoteCard = remoteCard
+	err := activeCard.RemoteCard.ConnectToCard(pairingCard.GetName())
 	if err != nil {
```
maevyn11, line 40: I don't love this pattern of manually setting RemoteCard on activeCard. It's not clear to the caller that this is how you establish a remote connection with another card and it seems likely to lead to accidental panics when a RemoteCard is not properly initialized. 

Feels like this is ripe for an interface that unifies LocalCounterParty and JumpboxCounterparty and provides a method or two on session to connect to either one. Thoughts?

repl/cardPairing.go
```diff
@@ -31,14 +31,14 @@ func cardPairLocal(c *ishell.Context) {
...
-		c.Println("pairing failed: ", err)
-		return
+		c.Err(err)
```
maevyn11, line 41: Needs a return afterwards so that the "successfully paired" message doesn't print when there's an error. Also does c.Err print the message?

session/remoteClient.go
```diff
@@ -1,4 +1,4 @@
-package remote
+package session
```
maevyn11, line 2: guessing this package moved for similar reasons as the others? Not sure it belongs in session.

maevyn11 (changes requested):

npmaile:
Opened a new pr because changes were going to be too divergent.

PR #89, opened 2022-01-28
Phonon deposit

maevyn11:


gui/api.go
```diff
@@ -108,6 +109,59 @@ func (apiSession apiSession) createPhonon(w http.ResponseWriter, r *http.Request
...
+	sess, err := apiSession.sessionFromMuxVars(vars)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusNotFound)
+		return
+	}
+	var depositPhononReq struct {
+		CurrencyType  model.CurrencyType
+		Denominations []int
+	}
+	reqBody, err := ioutil.ReadAll(r.Body)
```
npmaile, line 123: You can kill a couple of lines here by making a json decoder off of the body directly and use that. 

err := json.NewDecoder(r.Body).Decode(&denoms)
    if err != nil {
        http.Error(w, err.Error(), http.StatusBadRequest)
        return
    }

chain/chain.go
```diff
@@ -0,0 +1,29 @@
...
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
+)
+
+var ErrMissingPubKey = errors.New("phonon missing pubKey")
+var ErrMissingKeyIndex = errors.New("phonon missing KeyIndex")
+var ErrUnknownCurrencyType = errors.New("unknown currency type")
+
+/*Check the phonon's currency type and public key and returns a chain specific
+address as a hexstring*/
+func DeriveAddress(p *model.Phonon) (address string, err error) {
```
npmaile, line 17: I'd rather the DeriveAddress be an interface type, and you just call it on an ethereum provider or a bitcoin provider or whichever one. 

That way it's a little easier to test the calling functions because you can put in a test struct type that just has a dummy implementation instead of having a separate case for it.

npmaile (changes requested):

gui/api.go
maevyn11: Neat I like that.

chain/chain.go
maevyn11: yea, this could use some work. You could make it an interface type but then you'd still need to have the switch on currencyType that refers to the interface somewhere.  

I think it should maybe be a method on CurrencyType? That would be an easy place to collect chain specific stuff so there's a clean interface for external contributions. 

I guess I'm talking myself into an interface and just having a central method that switches which implementation is used based on the currencyType. Will change this.

gui/api.go
maevyn11: Suggestion implemented.

chain/chain.go
maevyn11: There's a wonky version of the interface in there now. Still need to figure out where to initialize it and exactly how the rest of the phonon data structure utilizes it.

npmaile (approved):

PR #93, opened 2022-02-16
Basic caching

npmaile:
this should implement a basic caching scheme to make card operations quite a bit faster.

### DIFF
--- a/gui/api.go
+++ b/gui/api.go
@@ -38,7 +38,6 @@ type apiSession struct {
 }
 
 type sessionCache struct {
-	// scandalous! a pointer to a bool! that's using 8 times more space than necessary
 	cachePopulated bool
 	phonons        map[uint16]*model.Phonon
 }
--- a/gui/api.go
+++ b/gui/api.go
@@ -421,12 +421,12 @@ func (apiSession apiSession) setDescriptor(w http.ResponseWriter, r *http.Reques
 		CurrencyType: model.CurrencyType(inputs.CurrencyType),
 	}
 	p.KeyIndex = uint16(index)
-	cache[sess.GetName()].phonons[p.KeyIndex] = p
 	err = sess.SetDescriptor(p)
 	if err != nil {
 		http.Error(w, "Unable to set descriptor", http.StatusBadRequest)
 		return
 	}
+	cache[sess.GetName()].phonons[p.KeyIndex] = p
 }
 
 func (apiSession apiSession) send(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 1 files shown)


gui/api.go
```diff
@@ -314,19 +347,32 @@ func (apiSession apiSession) listPhonons(w http.ResponseWriter, r *http.Request)
...
-	if err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
-	}
-	for _, p := range phonons {
-		p.PubKey, err = sess.GetPhononPubKey(p.KeyIndex)
+
+	phonons := []*model.Phonon{}
+	if cache[sess.GetName()].cachePopulated {
+		phonons = []*model.Phonon{}
```
maevyn11, line 353: Where does cachePopulated get set? I don't see it in createPhonon, initDeposit, or finalizeDeposit.

gui/api.go
```diff
@@ -314,19 +347,32 @@ func (apiSession apiSession) listPhonons(w http.ResponseWriter, r *http.Request)
...
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+		}
+		for _, phonon := range phonons {
+			cache[sess.GetName()].phonons[phonon.KeyIndex] = phonon
+		}
+		cache[sess.GetName()].cachePopulated = true
+	}
```
maevyn11, line 375: OK, I see. So this just caches lists after the first one.

gui/api.go
```diff
@@ -375,7 +421,7 @@ func (apiSession apiSession) setDescriptor(w http.ResponseWriter, r *http.Reques
 		CurrencyType: model.CurrencyType(inputs.CurrencyType),
 	}
 	p.KeyIndex = uint16(index)
-
+	cache[sess.GetName()].phonons[p.KeyIndex] = p
 	err = sess.SetDescriptor(p)
```
maevyn11, line 425: seems like this should maybe be after the error is checked on `err = sess.SetDescriptor`. If that fails then the cache and card will be out of sync for this keyIndex.

gui/api.go
```diff
@@ -400,7 +446,10 @@ func (apiSession apiSession) send(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "Unable to convert index to int:"+err.Error(), http.StatusBadRequest)
 		return
 	}
+
 	err = sess.SendPhonons([]uint16{uint16(index)})
+	delete(cache[sess.GetName()].phonons, uint16(index))
+
```
maevyn11, line 452: again this should probably be after the error check. If the key fails to send the cache will be out of sync.

maevyn11 (changes requested):
Looks good for the most part. Just a couple small comments that I think will make it more robust.

gui/api.go
npmaile: yeah, the idea is that the cache isn't fully populated until a full list has been done at least one time. After that anything that would modify the cache is modified in the cache along with the on card change happening

gui/api.go
npmaile: fixed

gui/api.go
npmaile: fixed

maevyn11 (approved):
LOOKS GOOD

PR #97, opened 2022-03-09
Separate connect/pair. Card Mutex. Synchronus request to session object from remote connection. 

npmaile:
Nearly complete. Needs testing and a completion of the localcounterparty


model/sessionRequest.go
```diff
@@ -0,0 +1,132 @@
+package model
+
```
maevyn11, line 2: OK, so this is like an interface definition of the actions that a counterparty can request of a session it has connected to? 

And the requests are identified by a string returned by GetName() which matches the struct name, where the handleRequest method in session uses a switch statement to determine how each request is handled? 

Just posing questions to make sure I understand the design.

orchestrator/session.go
```diff
@@ -278,20 +345,79 @@ func (s *Session) GenerateInvoice() ([]byte, error) {
...
 
+func (s *Session) ConnectToRemoteProvider(RemoteURL string) error {
+	u, err := url.Parse(RemoteURL)
+	if err != nil {
+		return fmt.Errorf("unable to parse url for card connection: %s", err.Error())
+	}
+	log.Info("connecting")
+	//guard
+	remConn, err := remote.Connect(s.remoteMessageChan, fmt.Sprintf("https://%s/phonon", u.Host), true)
+	if err != nil {
```
maevyn11, line 376: Ok, so this s.remoteMessageChan arg is how the remote connection communicates with the local session. Took me a minute to find how the handleRequest loop was being contacted. 

Will this work correctly if a user tries to connect to multiple cards within a single session? For example, pairing with card A, doing a transaction, and then pairing with card B to do a transaction. I suspect the s.remoteMessageChan should only receive messages from one card at a time so as long as that's the case it should be alright? Wondering if the downstream steps like ConnectToCounterparty and Jumpbox have been tested in this case? Not critical at this stage but something to consider moving forwards.

orchestrator/terminal.go
```diff
@@ -2,37 +2,33 @@ package orchestrator
...
-	sessions []*session.Session
+	sessions []*Session
 }
 
 type remoteSession struct {
 	counterParty model.CounterpartyPhononCard
 }
 
 var ErrRemoteNotPaired error = errors.New("no remote card paired")
+var ErrNoSession error = errors.New("No connected session with id found")
```
maevyn11, line 21: caps in error message

orchestrator/session.go
```diff
@@ -55,10 +70,23 @@ func NewSession(storage model.PhononCard) (s *Session, err error) {
 		log.Error("could not run session connect: ", err)
 		return nil, err
 	}
-
+	// launch session request handler
+	go s.handleIncommingSessionRequests()
```
maevyn11, line 74: I was skeptical at first but after reading through most of the code I think having a request handler loop in session actually makes a lot of sense. My concern is mainly that I worry about coupling Session with the implementation details of Jumpbox, therefore making it hard to implement other network providers like local and decentralized p2p, but I think the design here sufficiently separates the details so that concerns can be fairly easily separated when we need to.   

Also, typo here, too many mm's in incoming.

orchestrator/session.go
```diff
@@ -55,10 +70,23 @@ func NewSession(storage model.PhononCard) (s *Session, err error) {
...
 	return s, nil
 }
 
+// loop until killed
+func (s *Session) handleIncommingSessionRequests() {
+	for {
+		select {
+		case req := <-s.remoteMessageChan:
+			s.handleRequest(req)
+		case <-s.remoteMessageKillChan:
```
maevyn11, line 84: What's the idea with the remoteMessageKillChan? I don't see it being used anywhere atm.

remote/v1/message.go
```diff
@@ -1,4 +1,4 @@
-package remote
+package v1
```
maevyn11, line 2: v1 here presumably so Session could easily swap out the v1 client for an entirely different v2 implementation in future? Is the idea here that we might want to enable both clients to be an option so this is better than just versioning the code?

orchestrator/session.go
```diff
@@ -1,31 +1,41 @@
-package session
+package orchestrator
```
maevyn11, line 1: Soooo...I'll bite. Why did session move to orchestrator? I suppose I don't see a problem with it, just curious.

remote/v1/client/client.go
```diff
@@ -45,10 +46,101 @@ type RemoteConnection struct {
 	phononAckChan chan bool
 }
 
-// this will go someplace, I swear
 var ErrTimeout = errors.New("Timeout")
 
-func Connect(s *session.Session, url string, ignoreTLS bool) (*RemoteConnection, error) {
+// Requests into the card session
```
maevyn11, line 51: Do these calls make requests of the local session or of the remote session?

orchestrator/session.go
```diff
@@ -388,23 +515,84 @@ func (s *Session) FinalizeDepositPhonon(dc DepositConfirmation) error {
...
-// 		err = s.SetDescriptor(p)
-// 		if err != nil {
-// 			//TODO: work out what to do in the event that only some descriptors fail
-// 			log.Error("unable to set descriptor on deposited phonon at KeyIndex: %v. err: ", p.KeyIndex, err)
-// 		}
-// 	}
-// 	return phonons, nil
-// }
+// the panics are paths that should NEVER be found in runtime as it's already been determined by the case statement.
+func (s *Session) handleRequest(r model.SessionRequest) {
```
maevyn11, line 519: I don't have a good solution for it but it feels like this logic is being redefined several times throughout the program. Here, in remote client, and in remote server. It feels like we should be able to abstract that away somehow, though we're probably too far down this path to do it now. Just food for thought.

maevyn11 (changes requested):
Overall I think the PR looks good. Mostly fiddly formatting comments and clarifying questions from me. Of course also waiting for local counterparty to be implemented before final approval.

orchestrator/session.go
```diff
@@ -1,31 +1,41 @@
...
+	remoteMessageChan     chan (model.SessionRequest)
+	remoteMessageKillChan chan interface{}
+	active                bool
+	pinInitialized        bool
+	terminalPaired        bool
+	pinVerified           bool
+	Cert                  *cert.CardCertificate
+	name                  string
+	ElementUsageMtex      sync.Mutex
 }
```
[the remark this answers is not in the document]
npmaile: i wrote that line shortly after a Eureka! moment on how the app should be structured in the future given that it will run on things with secure enclaves other than the smart card.

orchestrator/session.go
npmaile: i just felt weird writing an infinite loop with no way of killing it. the channel will be used when we come up with a mechanism for deleting sessions

orchestrator/session.go
```diff
@@ -47,16 +46,23 @@ var ErrCardNotPairedToCard = errors.New("card not paired with any other card")
 func NewSession(storage model.PhononCard) (s *Session, err error) {
 	s = &Session{
 		cs:                    storage,
+		RemoteCard:            nil,
+		identityPubKey:        &ecdsa.PublicKey{},
```
maevyn11, line 50: Shouldn't this zero as nil instead of &ecdsa.PublicKey? Wondering if this is another editor autofill issue.

orchestrator/session.go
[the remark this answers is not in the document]
maevyn11: OK, I get the secure element reference now. Still find the name awkward though, I think because Usage is redundant with the fact that it's a lock. What about SecElementMtx?

remote/v1/client/client.go
```diff
@@ -53,6 +54,7 @@ func (c *RemoteConnection) getLocalCertificate() (*cert.CardCertificate, error)
 	req := &model.RequestCertificate{
 		Ret: make(chan model.ResponseCertificate),
 	}
+	c.logger.Debug("Requesting local card certificate")
 	c.sessionRequestChan <- req
```
maevyn11, line 58: logs should avoid capitalization too.

maevyn11 (changes requested):
Few small comments. Still needs a local counterparty implementation.

cmd/JumpBox.go
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 package cmd
 
 import (
-	"github.com/GridPlus/phonon-client/remote"
+	remote "github.com/GridPlus/phonon-client/remote/v1/server"
 	"github.com/spf13/cobra"
```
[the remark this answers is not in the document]
npmaile: for easy replacement if we need to do so. Also, there will likely be other remote clients for other communication mediums other than go -> go http/2

orchestrator/session.go
npmaile: the phonon terminal being in a different package than the terminal didn't work great for facilitating communication between sessions with the separate pair/connect steps

remote/v1/client/client.go
npmaile: into the local session. That way we have clear frontend -> terminal -> session -> remote ownership heirarchy instead of having the remote and session be peers. The remote can politely ask for requests of the session it's connected to instead of having control.

orchestrator/session.go
npmaile: It's deceptively similar, but ultimately different

remote/v1/client/client.go
npmaile: will fix across the entire repo in a later pr. There's mistakes in other files and I don't want the pr to include them. it's a fairly simple regex to do it, so it'll take approx 2 minutes to do.

orchestrator/session_test.go
```diff
@@ -0,0 +1,108 @@
...
+	err := sess2.SendPhonons([]uint16{
+		0,
+	})
+	if err != nil {
+		t.Error(err.Error())
+	}
+
+}
+
+func TestE2ELocalSendPhonon(t *testing.T) {
```
maevyn11, line 68: Seems like there are some issues running this test for local counterparty. Also get an error trying to run cmd/pairCardToCard with two fresh cards installed with alpha applets.  `go run main/phonon.go pairCardToCard -r 0 -s 1`

repl/cardPairing.go
```diff
@@ -33,11 +32,14 @@ func cardPairLocal(c *ishell.Context) {
...
 	c.Println("starting pairing with ", pairingCard.GetName())
-	remoteCard := orchestrator.NewLocalCounterParty(pairingCard)
-
-	err := activeCard.PairWithRemoteCard(remoteCard)
+	err := activeCard.ConnectToLocalProvider()
+	if err != nil {
+		c.Printf("Error occured in pairing: %s", err.Error())
+		return
+	}
+	err = activeCard.ConnectToCounterparty(otherCards[selection].GetName())
```
maevyn11, line 40: getting an error trying to pair two alpha cards in the repl. 
```
starting pairing with  04ee191cc71fd218
{"level":"info","msg":"returning error from ConnectRemoteSession","time":"2022-03-22T12:04:14-05:00"}
Error occured in pairing process: counterparty card not found
```

orchestrator/session.go
maevyn11: ok, works as a placeholder for now then.

maevyn11 (changes requested):
Having some issues with the localCounterParty functionality. Haven't been able to get the tests, REPL, or cmd pairCardToCard command to work on my machine with two alpha cards. There are also a few minor outstanding comments to address. Once these are resolved we should be good to merge this in with the current master.

orchestrator/session_test.go
npmaile: test is functioning now.

repl/cardPairing.go
npmaile: fixed.

orchestrator/session_test.go
npmaile: fixed pairCardToCard command.

orchestrator/session.go
npmaile: done.

repl/cardPairing.go
maevyn11: nice, works great now. Only thing left to change is setting PairCounterparty to pairCounterparty and showing what the argument is supposed to be in the repl help text

maevyn11 (approved):
Looks good!

PR #101, opened 2022-03-11
Phonon redeem

maevyn11:
We'll have to merge one or the other of the active PRs before these can both go in. 
Major Features: 

- New chain package with mostly complete chainService interface
- Working ETH redeem with handling for chainIDs (ready to be extended to EVM compatible chains and forked or copied for ERC-721 support)
- Working InitDeposit API endpoint which integrates with frontend implemented metamask based deposits.
- MultiChainRouter to handle abstracting blockchain functions over different chain designs. 
- Medium term hack to use ExtendedTLVs to store ChainID on card and then parse it into a named struct field
- big.Int handling for NewDenomination. JSON string conversion into and out of Denominations that supports big.Int sized string numbers.


chain/ethChainService.go
```diff
@@ -0,0 +1,218 @@
+package chain
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"errors"
+	"math/big"
+
+	// "net/http"
```
npmaile, line 9: remove commented imports

chain/ethChainService.go
```diff
@@ -0,0 +1,218 @@
...
+	err = eth.cl.SendTransaction(ctx, signedTx)
+	if err != nil {
+		log.Error("error sending transaction: ", err)
+		return signedTx, err
+	}
+	log.Debug("sent redeem transaction")
+	return signedTx, nil
+}
+
+//London Signing Code
```
npmaile, line 190: I assume this will be used in a future iteration, but isn't necessary right now. Maybe should go on a branch instead of sitting commented out.

chain/ethChainService.go
```diff
@@ -0,0 +1,218 @@
...
+	log "github.com/sirupsen/logrus"
+)
+
+type EthChainService struct {
+	gasLimit  uint64
+	cl        *ethclient.Client
+	clChainID int
+}
+
+func NewEthChainService() (*EthChainService, error) {
```
npmaile, line 26: mild nitpick, but should return the interface type instead of a pointer to struct type unless there is eth specific exported functions to be used.

chain/multiChainRouter.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+//InitChainService and LoadChainService enable dynamic loading of different chain providers
+
+//Detect the chain service that should be used for each phonon based on CurrencyType
+func InitChainService(cur model.CurrencyType) (ChainService, error) {
```
npmaile, line 38: possibly store these in a static map someplace instead of a long switch case.

config/config.go
```diff
@@ -3,24 +3,64 @@ package config
 import (
 	"github.com/GridPlus/phonon-client/cert"
 	log "github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
```
npmaile, line 6: viper will likely work for now, but as we get more into trying to make this friendly for less technical users, we may wish to have configurations set in the app that are persisted across runs. I'm not exactly sure of libraries to support this, and we may be able to leverage viper anyway despite not being exaaaaactly the normal viper use case.

gui/api.go
```diff
@@ -346,7 +472,7 @@ func (apiSession apiSession) sessionFromMuxVars(p map[string]string) (*session.S
 	sessionName, ok := p["sessionID"]
 	if !ok {
 		fmt.Println("unable to find session")
-		return nil, fmt.Errorf("Unable to find sesion")
+		return nil, fmt.Errorf("unable to find sesion")
```
npmaile, line 475: s/sesion/session

gui/api.go
```diff
@@ -357,7 +483,7 @@ func (apiSession apiSession) sessionFromMuxVars(p map[string]string) (*session.S
 		}
 	}
 	if targetSession == nil {
-		return nil, fmt.Errorf("Unable to find sesion")
+		return nil, fmt.Errorf("unable to find sesion")
```
npmaile, line 486: s/sesion/session/

model/phonon.go
```diff
@@ -20,39 +24,79 @@ type Phonon struct {
...
+		return err
+	}
+
+	p.Address = phJSON.Address
+	p.AddressType = phJSON.AddressType
+	p.SchemaVersion = phJSON.SchemaVersion
+	p.ExtendedSchemaVersion = phJSON.ExtendedSchemaVersion
+	p.Denomination = phJSON.Denomination
+	p.CurrencyType = CurrencyType(phJSON.CurrencyType)
+	//idk if this is good
```
npmaile, line 84: Be more  confident in yourself!

phonon.yml
```diff
@@ -0,0 +1,3 @@
+#Sample Config File (Fill in values and store in $HOME/.phonon/phonon.yml)
```
npmaile, line 1: should search $XDG_CONFIG_HOME/.phonon/phonon.yml followed by $HOME/.phonon/phonon.yml followed by /usr/var/phonon/phonon.yml at least on linux systems
we need to figure out what to do on windows systems as well.

chain/ethChainService.go
```diff
@@ -0,0 +1,218 @@
...
+	//Collect on chain details for redeem
+	nonce, onChainBalance, suggestedGasPrice, err := eth.fetchPreTransactionInfo(ctx, common.HexToAddress(p.Address))
+	if err != nil {
+		return "", err
+	}
+	redeemValue := eth.calcRedemptionValue(onChainBalance, suggestedGasPrice)
+	log.Debug("transaction redemption value is: ", redeemValue)
+
+	//If gas would cost more than the value in the phonon, return error
+	if suggestedGasPrice.Cmp(onChainBalance) != -1 {
```
npmaile, line 63: This check should be done before destruction of the phonon on-card using the public key to prevent people from destroying phonons they won't be able to redeem and ending up with a private key they have to store someplace.

npmaile (changes requested):

chain/ethChainService.go
maevyn11: done

chain/ethChainService.go
maevyn11: yea, possibly used in future, but not useful enough code to be worth keeping around. Removed it.

chain/multiChainRouter.go
maevyn11: Yea, realized that map exists as mcr.chainServices and it's being built already with InitBuiltinChainServices. 

Think I was getting ahead of myself with the lazy loading functions imagining that 3rd party implementations would be pluggable. Since that's not being implemented here, just going to remove that code in favor of the simple initialization of all supported chainServices on startup.

chain/ethChainService.go
maevyn11: Yea, I agree. I thought of that as I was working through this as well. :( 
  
Unfortunately, that's pretty hard to do within this paradigm, because we can't call back to session from within this method to get that data mid flow. I guess we would have to break up RedeemPhonon into multiple steps that get orchestrated by Session, like ValidateRedeemReady and then RedeemPhonon. This should probably also account for sending info back to the frontend so the user can view things like gas price and accept them before executing the redeem. 

Since this would entail significant additional coding which we don't necessarily need for testnet, I suggest we create an issue for a v2 of this interface definition that improves upon the full user flow and defer this work till then.

config/config.go
maevyn11: yea agreed we'll want to support that. I believe viper does actually have good support for not just reading but also writing config files, so things are saved across runs.

gui/api.go
maevyn11: done

gui/api.go
maevyn11: done

phonon.yml
maevyn11: https://github.com/GridPlus/phonon-client/issues/103

chain/ethChainService.go
maevyn11: https://github.com/GridPlus/phonon-client/issues/104

PR #112, opened 2022-03-28
ERC721 and ERC20 Currency Types

soysenor:
REPL
- Added `VerifyBalance` method.  Takes `keyIndex` for Phonon, then routes through multiChainRouter to ETHChainService, and switches between `checkERC721Balance`, `checkETHBalance`, `checkERC20Balance` pending Currency Type with support from helper function `checkFetchFromAddress`.  Go-bindings for ERC721 and ERC20 contract are currently sourced from one of my repos, but should be moved into phonon. 
- Incorporated Tagging into `SetDescriptor`.  Gets required tags for ERC721 and ERC20 (see `CurrencyTypeTagsRequired` map on model/phonon.go), prompts user for inputs (`TagPhononContractAddress` and `TagPhononContractTokenID` - see card/apdu too), and passes user inputs `[]PhononTag` to `AddTags` method on Phonon model to be structured according to TLV standard.
- Added `ChainID` as a mandatory 4th parameter for `SetDescriptor`.

GUI
- Updated `initDepositPhonons` to take `Tags` as slices of key value strings (see `[][]PhononTag`), which are parsed into TLV standard via `AddTags` method in `initDepositPhonons` in session/session.   But because `setDescriptor` isn't processed until `FinalizeDepositPhonons`, I had to work through `UnmarshalJSON` and `MarshalJSON` on model/phonon for ExtendedTLV to pass data back and forth between web and client.  I'm sure you'll have a more sophisticated approach to this, but I made it work for my tags (see `marshalTags`).  

Additional Notes:
- Currently the following work for ERC721 and ERC20:  `create`, `setDescriptor`, `initDepositPhonons`, `verifyBalance`.
- Not fully integrated: `redeem`.   You'll see I updated the `RedeemPhonon` method for EthChainService  to switch on currencyType; `redeemETH` and `redeemToken`.  This hasn't been fully tested yet.  Is a little tricky with the bug in the applet limiting currency types, and with the react app actively being worked on.  But it's close so I'm incorporating the code in anyway.  Also, I would not call redemption for ERC20 nor ERC721 fully integrated until the pre-requisite transfer of ETH to the phonon is thought out; this needs to be done in the UI.
- Added CurrencyType `EthereumERC721` and `EthereumERC20` to model/phonon.
- Known Bug: I need to talk to you about a bug I am seeing when attempting to `setDescriptor` for ERC721.   I had no issues assigning both tags, until recent updates to Master.  What I'm finding is I am able to assign my `TagPhononContractAddress` OR `TagPhononContractTokenID` tags successfully, but not BOTH at the same time, when the length of `TagPhononContractAddress` > 39.  This of course doesn't work for ERC721 since contract addresses are 42 characters, but you can see adding both tags works for ERC721 if you shorten the contract address you input.  I am probably doing something wrong, but need help understanding what's going on.

### DIFF
--- a/chain/ethChainService.go
+++ b/chain/ethChainService.go
@@ -6,22 +6,31 @@ import (
 	"errors"
 	"math/big"
 
+	"golang.org/x/crypto/sha3"
+
 	"github.com/GridPlus/phonon-client/model"
 	"github.com/GridPlus/phonon-client/util"
+	"github.com/GridPlus/phonon-client/tlv"
+	"github.com/GridPlus/phonon-client/card"
+
 	"github.com/ethereum/go-ethereum"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
-
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/params"
 	"github.com/ethereum/go-ethereum/core/types"
 	ethcrypto "github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/ethclient"
+
 	log "github.com/sirupsen/logrus"
+	erc721 "github.com/soysenor/eth-go-bindings/erc721"
+	erc20 "github.com/soysenor/eth-go-bindings/erc20"
 )
 
 //Composite interface supporting all needed EVM RPC calls
 type EthChainInterface interface {
 	bind.ContractTransactor
 	ethereum.ChainStateReader
+	bind.ContractBackend
 }
 type EthChainService struct {
 	gasLimit  uint64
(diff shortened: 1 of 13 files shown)


repl/phonons.go
```diff
@@ -102,11 +107,41 @@ func setDescriptor(c *ishell.Context) {
...
 	c.Println("setting descriptor with values: ", uint16(keyIndex), currencyType, denomination)
 	p := &model.Phonon{
 		KeyIndex:     uint16(keyIndex),
 		CurrencyType: currencyType,
 		Denomination: denomination,
+		ChainID: chainId,
+	}
+
+	// Capture Required Tag Fields
+	requiredTags := model.CurrencyTypeTagsRequired[currencyType]
```
npmaile, line 126: @maevyn11 I'm not sure what needs to be done with the tags

repl/phonons.go
```diff
@@ -102,11 +107,41 @@ func setDescriptor(c *ishell.Context) {
...
+	newTags := make([]model.PhononTag, 0)
+	if len(requiredTags) > 0 {
+		c.Println("Additional tags are required.")
+
+		for i, tag := range requiredTags {
+			c.Printf("Input %v: ", tag)
```
npmaile, line 132: %+v gives the field names as well, which may be useful

npmaile (changes requested):

chain/ethChainService.go
```diff
@@ -147,6 +195,118 @@ func (eth *EthChainService) dialRPCNode(chainID int) (err error) {
...
+	gasLimit, err := eth.cl.EstimateGas(ctx, ethereum.CallMsg{
+		To:   &contractAddress,
+		Data: data,
+	})
+	
+	if err != nil {
+		return "", errors.New("cannot estimate gaslimit for redemption")
+	}
+	leftOverValue := eth.calcRedemptionValue(onChainBalance, big.NewInt(int64(gasLimit)))
+	if leftOverValue < 0 {
```
maevyn11, line 295: This doesn't compile on my machine. I think with the *big.Int you'll need to use leftOverValue.Cmp() and check the output, unless there is some big.Int operator support in a newer version of golang I'm not aware of.

chain/ethChainService.go
```diff
@@ -197,3 +357,122 @@ func (eth *EthChainService) submitLegacyTransaction(ctx context.Context, nonce u
 	log.Debug("sent redeem transaction")
 	return signedTx, nil
 }
+
+func (eth *EthChainService) checkFetchFromAddress(p *model.Phonon) error {
```
maevyn11, line 361: I don't think this function is necessary, simply calling eth.DeriveAddress() will ensure that there is an address present. eth.DeriveAddress actually never returns an error, so the error check here isn't needed.

maevyn11 (changes requested):
Some initial feedback.

soysenor:
Postponing work until after testnet.

PR #123, opened 2022-04-01
Move session add

npmaile:
makes testing a little easier. Doesn't change anything other than the fact that manually generating a session from a cs will automatically add it to the terminal


maevyn11:
Closing in favor of https://github.com/GridPlus/phonon-client/pull/124

PR #129, opened 2022-04-13
Phonon priv key mock

maevyn11:


card/apdu.go
```diff
@@ -126,19 +126,14 @@ type Command struct {
...
-			return errors.New("unspecified error for command")
-		}
-	}
-	return ret
-
+	err, exists := cmd.PossibleErrs[int(res.Sw)]
+	if exists {
+		return err
+		//Return unspecified error if code is not listed in command and is < 0x9000
+	} else if res.Sw < SW_NO_ERROR {
```
npmaile, line 133: this won't work. There are errors above 9000 that can be returned.

npmaile (changes requested):

card/phononCommandSet.go
```diff
@@ -521,7 +521,7 @@ func (cs *PhononCommandSet) ListPhonons(currencyType model.CurrencyType, lessTha
 	log.Debugf("p2: % X", p2)
 	cmd := NewCommandListPhonons(0x00, p2, cmdData)
 	resp, err := cs.sc.Send(cmd)
-	if err != nil {
+	if err != nil && err != ErrDefault {
```
npmaile, line 524: need to explicitly check to make sure it's sw1 is 90. there could be another err that isn't specified thrown by some corner case that will break this

npmaile (changes requested):

card/phononCommandSet.go
maevyn11: checkContinuation a few lines down should handle this case. 
```
//Generally checks status, including extended responses
func checkContinuation(status uint16) (continues bool, err error) {
	if status == 0x9000 {
		return false, nil
	}
	if status > 0x9000 {
		return true, nil
	}
	return false, ErrUnknown
}
```

This once again codifies the >9000 is not an error pattern. Checking specifically for 0x90 would also codify that 0x90XX cannot be an error, though we know that is not a perfect description.  

I think the root issue here is that the card can return a range of response codes for a single state, which is "List returned successfully and has a continuation." I think the right solution for this is to modify the card code to always return 0x9001 when there is an extended list, rather than incrementing the value, since that increment is not useful on the client anyhow and cannot be processed sanely by HumanReadableErr as it is. 

Then we could specify 0x9001 as ErrListContinues and check for it specifically, and consolidate all of these error handling functions, cs.sc.Send, checkPhononTableErrors, and checkContinuation, into the CmdErrTable list to be handled the same as the more up to date commands. 

If this code functions correctly with the current tests, I propose we go with the current solution in this PR for now, make an issue out of my proposal above for the card and push it to the release after testnet.

npmaile (approved):

PR #134, opened 2022-04-28
Mock card native phonon support

maevyn11:


card/commandEncoding.go
```diff
@@ -194,6 +195,8 @@ func parseDestroyPhononResponse(resp []byte) (privKey *ecdsa.PrivateKey, err err
 	if err != nil {
 		return nil, err
 	}
+
+	fmt.Println("raw PrivateKey: " + string(rawPrivKey))
```
npmaile, line 199: make this a log debug command

npmaile (changes requested):
we gotta get rid of this abc.txt
that was my b

npmaile (approved):

PR #139, opened 2022-06-03
add frontend to build and compilation

npmaile:


maevyn11 (approved):
Nice, this is really slick.

PR #143, opened 2022-06-15
complete logs publishing

npmaile:


card/phononCommandSet.go
```diff
@@ -57,6 +57,7 @@ type PhononCommandSet struct {
 func NewPhononCommandSet(c types.Channel) *PhononCommandSet {
 	var err error
 	conf, err := config.LoadConfig()
+	fmt.Println("config loaded")
```
maevyn11, line 60: should probably remove this random Println. If it's needed should be a log statement.

config/config.go
```diff
@@ -14,6 +15,8 @@ type Config struct {
 	//PhononCommandSet
 	AppletCACert []byte //One of the CA certificates listed in the cert package. Used as default if Certificate not set
 	Certificate  string //string ID to select a certificate
+	// logg exporting
```
maevyn11, line 18: too many g's

hooks/logging.go
```diff
@@ -0,0 +1,66 @@
...
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		fmt.Println("unable to send logs to telemetry server", err.Error())
+		return err
+	}
+	if resp.StatusCode != http.StatusOK {
+		fmt.Println("unable to send logs to telemetry server", string(body))
+	}
+	return err
+}
```
maevyn11, line 62: So I'm guessing this implements the interface logrus looks for to ship the log remotely? Pretty cool.

go.mod
```diff
@@ -10,6 +10,8 @@ require (
 	github.com/btcsuite/btcutil v1.0.3-0.20201208143702-a53e38424cce
 	github.com/certusone/yubihsm-go v0.2.0
 	github.com/ebfe/scard v0.0.0-20190212122703-c3d1b1916a95
+	github.com/elastic/go-elasticsearch v0.0.0 // indirect
+	github.com/elastic/go-elasticsearch/v7 v7.17.1 // indirect
```
maevyn11, line 14: Can these dependencies be removed since we're not using elasticsearch? Seems like it's bringing in a bunch of transitive dependencies we may not need.

config/config.go
```diff
@@ -56,6 +59,12 @@ func LoadConfig() (config Config, err error) {
 	if err != nil {
 		return DefaultConfig(), err
 	}
+	// Possibly not the best place to put this, but it does a good job of setting this up before an interactive session
+	if config.TelemetryKey != "" {
+		log.Debug("setting up logging hook")
+		log.AddHook(hooks.NewLoggingHook(config.TelemetryKey))
+	}
+
```
maevyn11, line 67: I had the same dilemma about keeping the config.Certificate check below here. I think it's fine for now, though maybe it's more correct to push this stuff into relevant modules as it builds up.

maevyn11 (changes requested):
Just a couple small changes and a hopeful request to remove some unneeded dependencies.

hooks/logging.go
npmaile: yes

config/config.go
npmaile: lit

go.mod
npmaile: go mod tidy took care of this

card/phononCommandSet.go
npmaile: delet'd

config/config.go
```diff
@@ -61,7 +61,7 @@ func LoadConfig() (config Config, err error) {
 	}
 	// Possibly not the best place to put this, but it does a good job of setting this up before an interactive session
 	if config.TelemetryKey != "" {
-		log.Debug("setting up logging hook")
+		log.Debug("setting up loging hook")
```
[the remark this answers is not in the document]
maevyn11: well, 3 g's in logging technically, but you get the idea.

Issue #151, opened 2022-07-05
Implement Init API Route


maevyn11:
In order for the frontend to be able to initialize a user's card, a route in api.go needs to be implemented that allows for INIT'ing a card and then subsequently using it. 

Some extra care may be needed to fix, handle, or workaround any issues caused by https://github.com/GridPlus/phonon-client/issues/142 or https://github.com/GridPlus/phonon-client/issues/131

PR #156, opened 2022-07-07
Init route

maevyn11:
Added InitRoute and updated ListSessions to return sessions status including whether the Pin is initialized.

Closes https://github.com/GridPlus/phonon-client/issues/151


gui/swagger.yaml
```diff
@@ -27,17 +27,40 @@ paths:
       produces:
       - "application/json"
       responses:
-        "404":
```
npmaile, line 30: not sure why no cards being plugged in is a 500 error

npmaile (changes requested):

gui/swagger.yaml
maevyn11: idk, seemed like a 404 page not found was misleading since the page is there, it's just it can't serve the expected response because there's no card plugged in behind the scenes. My thought is it's a 500 because the server can't handle the request given it's state. Neither one really seems right but I wasn't sure what else to use.

maevyn11: OK, guess that's my misconception then, 404 is just as good so no reason to change it. Fixed https://github.com/GridPlus/phonon-client/pull/156/commits/9252c0b78a417633cc7506e4a6d124ad9893fcb7

npmaile (changes requested):

npmaile (approved):

douglance (approved):

PR #160, opened 2022-07-11
add browser open also go mod tidy also make linter happy

npmaile:
confirmed working as expected on windows, mac, and archlinux


maevyn11 (approved):
Looks good to me, works on my machine.

PR #166, opened 2022-07-18
initial implementation of BER-TLV parser

MSalopek:
Not at all properly tested.

Instead of using the current version, this allows specifying VALUE length > 256 (currently unbounded, but has to be < 65535).


tlv/tlv.go
```diff
@@ -86,6 +87,105 @@ func ParseTLVPacket(data []byte, constructedTags ...byte) (TLVCollection, error)
...
+
+	// length group ob bytes starts at b[1] and can be from 1 to 3 bytes long
+	if b[1] <= 0x7F {
+		return tag, int(b[1]), 2, nil
+	}
+
+	// first 2 bytes of group represent length
+	// used for TLV.Value lengths between 128 and 255
+	if b[0] == 0x81 {
+		if len(b)-1 <= 0 {
```
maevyn11, line 169: would it be just as correct to code it this way? 
```
if len(b) != 2 {
return err...
}
``` 
If so I think it's a bit more clear what the check is. Took me a minute to figure out what this was doing.

tlv/tlv.go
```diff
@@ -86,6 +87,105 @@ func ParseTLVPacket(data []byte, constructedTags ...byte) (TLVCollection, error)
...
+			return 0, 0, 0, errors.New("expected 2 bytes encoding - missing second byte")
+		}
+
+		return tag, int(b[2]), 3, nil
+	}
+
+	// first 3 bytes of group represent length
+	// used for TLV.Value lengths between 256 - 65535
+	if b[0] == 0x82 {
+		if len(b)-2 <= 0 {
```
maevyn11, line 179: Same question as on the above check. 
```
if len(b) != 3
```

tlv/tlv.go
```diff
@@ -86,6 +87,105 @@ func ParseTLVPacket(data []byte, constructedTags ...byte) (TLVCollection, error)
...
+
+	tag := b[0]
+
+	// length group ob bytes starts at b[1] and can be from 1 to 3 bytes long
+	if b[1] <= 0x7F {
+		return tag, int(b[1]), 2, nil
+	}
+
+	// first 2 bytes of group represent length
+	// used for TLV.Value lengths between 128 and 255
```
maevyn11, line 167: thanks for the comments, very helpful.

tlv/tlv.go
maevyn11: Also, how does this check work when the data passed in is the entire payload for a command. Since it may consist of a group of TLVs, couldn't the length of b be anything? Or am I missing something?

tlv/tlv.go
```diff
@@ -86,6 +87,105 @@ func ParseTLVPacket(data []byte, constructedTags ...byte) (TLVCollection, error)
...
+	// tag with length 0
+	// TODO: check correctness
+	if len(data) < 2 {
+		return result, nil
+	}
+
+	tag, length, seek, err := parsePrefix(data)
+	if err != nil {
+		return result, err
+	}
```
maevyn11, line 126: Nice, I like the new approach with parsePrefix and then seeking to read the value. Very clean. More readable than my old looping implementation above. Probably more efficient too.

tlv/tlv_test.go
```diff
@@ -61,3 +62,22 @@ func TestParseTLV(t *testing.T) {
...
 }
+
+// Example with tag being 0x11
+// A data length of 64 bytes will be 0x40 - so it would look like 0x1140
+// A data length of 130 bytes will be 0x81 0x83 - so it would look like 0x118183
+// A data length of 280 bytes will be 0x82 0x01 0x18 - so it would look like 0x11820118
+func TestParseBERTLVPacket(t *testing.T) {
+	cases := []string{
+		"A4 11 0C 06 4D 7E 6C 6C 65 72 02 01 1E 01 01 00 80 01 00",
+	}
```
maevyn11, line 73: Guessing you had this in mind for later but we'll want to check a bunch of additional cases, including using constructedTags, packets at and in between each length boundary, packets that exceed the max size extended, and invalid packets.

maevyn11: The card implementation is being developed by @SecureElement at the same time, so we can get some sample output from the new extended BERTLV implementation there when it's ready and plug it in here to make sure they're in accordance.

maevyn11 (changes requested):
Hey overall looks really good. Think you're on the right track. Left a couple questions on the error handling in parsePrefix. Not sure if I spotted a potential bug or if I'm missing something about the implementation. Otherwise, mostly just suggestions to clarify a few things and for testing.

tlv/tlv_test.go
MSalopek: Having some examples would be awesome.

If we can get some soon, I'd like to use them for table tests.

There will be some basic sanity check tests that check if the prefix parsing is working as expected.

tlv/tlv.go
```diff
@@ -86,6 +87,105 @@ func ParseTLVPacket(data []byte, constructedTags ...byte) (TLVCollection, error)
...
+- case LENGTH < 0x7f (127) -> LENGTH encoded using single byte
+- CASE LENGTH > 0x7f (127) -> LENGTH encoded using 2 bytes
+    - the first byte will be 0x81
+	- second byte is the LENGTH
+- case LENGTH > 256 -> LENGTH encoded using 3 bytes
+    - first byte will be 0x82
+	- second and third bytes are the actual value LENGTH
+*/
+
+func ParseBERTLVPacket(data []byte, constructedTags ...byte) (TLVCollection, error) {
```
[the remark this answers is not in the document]
MSalopek: True, that's just for testing and showcasing development. If I change the name, all the tests in the code will fail, and all the parsing and TLV building will fail.

tlv/tlv.go
```diff
@@ -86,6 +87,105 @@ func ParseTLVPacket(data []byte, constructedTags ...byte) (TLVCollection, error)
...
+tag to find internal TLV's and append them to the collection as flattened entries
+
+Extends the max length of Value to 65535 instead of 256 bytes.
+
+The TLV is parsed as follows:
+TAG: 1st byte
+LENGTH: can be encoded using a GROUP 1 to 3 bytes
+- 1st byte of LENGTH group defines the number of bytes encoding the LENGTH
+- case LENGTH < 0x7f (127) -> LENGTH encoded using single byte
+- CASE LENGTH > 0x7f (127) -> LENGTH encoded using 2 bytes
```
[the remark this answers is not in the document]
MSalopek: That is correct. The comment is wrong. It should be as you stated (`case LENGTH > 0x7f (127) and LENGTH <= 0xFF (256)`)

tlv/tlv.go
MSalopek: There is a bug in the code, yes :D Thank you for finding it.

The func `parsePrefix` takes the full TLV and checks the start bytes because you cannot know how long the prefix is unless you peek into the slice - the `Tag-Length` part of TLV can be 2 to 4 bytes long.

Just slicing the TLV with data[0:4] would panic in case  len(TLV) < 4 - which is a valid TLV with only the TAG and a single LENGTH byte that should be set to 0  (TLV with 0 length value), or a TLV that has a single VALUE byte.

The initial intention was the following:
- check that the bytes defining the length are in the correct position from the start of the array
- if you subtract the byte position from the prefix length and end up with a negative value, something is wrong

The parsing was updated with regards to your comments.
```golang
func parsePrefix(b []byte) (byte, int, int, error) {
	if len(b) == 0 {
		return 0, 0, 0, errors.New("prefix not provided")
	}

	if len(b) < 2 {
		return 0, 0, 0, errors.New("prefix too short")
	}

	tag := b[0]

	if b[1] <= 0x7F {
		return tag, int(b[1]), 2, nil
	}

	if b[1] == 0x81 {
		if len(b) < 3 {
			// TLV too short to encode 2 length bytes (TAG+2 LENGTH bytes)
			return 0, 0, 0, errors.New("invalid TLV - expected 2 bytes length encoding")
		}
		return tag, int(b[2]), 3, nil
	}

	if b[1] == 0x82 {
		if len(b) < 4 {
			// TLV too short to encode 3 length bytes (TAG+3 LENGTH bytes)
			return 0, 0, 0, errors.New("invalid TLV - expected 3 bytes length encoding")
		}
		return tag, int(binary.BigEndian.Uint16(b[2:4])), 4, nil
	}

	return 0, 0, 0, errors.New("invalid value prefix encoding")
}
```

tlv/tlv.go
MSalopek: Added answer above. The function takes the full TLV as argument, so instead of checking inequality it checks if the overall length is less than required minimal length.

MSalopek:
It would be nice to have 5-6 examples for testing.

tlv/tlv.go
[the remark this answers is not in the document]
maevyn11: 👍

tlv/tlv.go
```diff
@@ -120,7 +120,7 @@ func ParseBERTLVPacket(data []byte, constructedTags ...byte) (TLVCollection, err
 		return result, nil
 	}
 
-	tag, length, seek, err := parsePrefix(data)
+	tag, length, seek, err := parsePrefix(data[0:4])
```
maevyn11, line 123: Possibly panic on the slice here if the packet is only 3 bytes long? (1 byte tag, 1 byte length, 1 byte value) could be a valid TLV. May have to put a guard above to check the length and pass in a smaller slice if the data is less than 4 bytes longs.

tlv/tlv.go
```diff
@@ -146,40 +146,41 @@ func ParseBERTLVPacket(data []byte, constructedTags ...byte) (TLVCollection, err
...
+		return 0, 0, 0, errors.New("prefix not provided")
 	}
 
 	if len(b) < 2 {
 		return 0, 0, 0, errors.New("prefix too short")
 	}
 
+	// NOTE: BER-TLV can contain multiple TAG bytes
+	// our implementation only uses single byte tags
+	// due to some smart contract restrictions
```
maevyn11, line 164: It's note exactly due to smart contract restrictions, but mainly to make the encoding more efficient for interacting with the resource constrained smart card.

tlv/tlv.go
```diff
@@ -146,40 +146,41 @@ func ParseBERTLVPacket(data []byte, constructedTags ...byte) (TLVCollection, err
...
-	if b[0] == 0x82 {
-		if len(b)-2 <= 0 {
-			return 0, 0, 0, errors.New("expected 3 bytes encoding - missing third byte")
+	if b[1] == 0x82 {
+		if len(b) < 4 {
+			// TLV too short to encode 3 length bytes (TAG+3 LENGTH bytes)
+			return 0, 0, 0, errors.New("invalid TLV - expected 3 bytes length encoding")
 		}
-
 		return tag, int(binary.BigEndian.Uint16(b[2:4])), 4, nil
```
maevyn11, line 184: May need to check on the maximum value that can be encoded here and ensure it's not more than the card can handle. I'm not exactly sure what the upper limit is. Maybe @SecureElement can give the exact value here when he's available. (He's out on vacation until Monday)

maevyn11 (changes requested):
Couple follow up comments on the changes

tlv/tlv.go
maevyn11: Think this does fix parsePrefix's byte logic though.

tlv/tlv.go
maevyn11: Nice, yes I think this fixes the issue. I would add a check at the top that len(b) <= 4, as otherwise we can't guarantee that the caller will always obey the rules and restrict the input length.

tlv/tlv.go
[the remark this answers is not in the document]
maevyn11: 👍

tlv/tlv_test.go
maevyn11: Here's an example APDU. An APDU always has a header of 4 bytes and then is followed by the data. Normally that data is TLV encoded but there are a few commands inherited from the upstream codebase which don't encode in TLV format. 
APDU:
```
0032000030D0582F4568171AC5A8375F100AE6B7CF951A4296E68B37EA6B33919528FDD7D42C16981DC2A14957AEC919F1E654922C
```
TLV (first 4 bytes removed): 
```
30D0582F4568171AC5A8375F100AE6B7CF951A4296E68B37EA6B33919528FDD7D42C16981DC2A14957AEC919F1E654922C
```

You can actually generate these yourself fairly easily. As long as the app's log level is Debug it will automatically generate an `apdu.log` file in the root of the project whenever the cli is invoked. This will include all the APDU's launched in that session, which you can manually parse the TLV's out of to generate test cases. 

An example of running the cli: 
```
go run main/phonon.go createPhonon 1
```

These aren't perfect for this as they are commands instead of responses, but they should contain valid TLV's for testing. We'll have to get some sample output from the new card changes to get some that are sufficiently long to test the extended length aspect, but the existing ones should work well for regression testing.

maevyn11: Let me know if you have any trouble generating these, not sure if it'll work right without a card available. We can generate some for you to use if so.

tlv/tlv.go
MSalopek: Meant to write "smart card", and wrote "smart contract". Too many things are smart these days :D

SecureElement:
> Not at all properly tested.
> 
> Instead of using the current version, this allows specifying VALUE length > 256 (currently unbounded, but has to be < 65535).

The maximum ( for now ) is 32K ( 0x7fff ) in Javacard.  We should max this a static variable that can be changed as needed.  RIght now,  it should be set to the maximum size that the secure channel can handle ( around 200 bytes ).

SecureElement:
unencrypted BerTLV sample 
80510000#(90819030900202000080410468F84EC5BF0B5F4EB94ED720EF99FC1EFA4FECE6EA9B9138C3F4C07E169C0940DBF9128BD9ABAB58E8B8AEAA9F6313AEB9D17B05E37201FD68201F4AEB675903304502205AA91139AA469435F5F5EE3D18E333276C97F39792D02300D1DB89D89853EC2002210090AD60F5767B1A62FD91054B48DCBAE5CEA7B469349993A4C2A38698D92A3BE9\
912096DC0119296C3C2236E5146A72BCBD284CE68535727A43C2B784B378CB243C21)
the pound sign signifies Lc where the script calculates the length of data being sent and places it there.  Will include a few more examples

SecureElement:
80530000#(93473045022100C7DC911449F799FAD2F2522D2CDFF88B69A4BFFABC4D90A734454BAA63CE28AE0220128B78E3C1EA07F1B9CB84D38DA8BF6EC64F107EE73ED78275EEF9A6691BB7E1)

SecureElement:
80500000#{908191309102020000804104D2B04AE13EE4D991A94065009DC7C0C48620663FB8B5C8F3890FF53F00CA7D5148BFAD218922E076E30AB8DC1509C7CF1296772EEAC37F8C7A7180920CD908503046022100FD12CCA7BFA3139FF383B655B6018811042D2654DF5C2F56A00B9951A67824A602210080F8C336A5DF1205FA1B732512126B4C0370BB41E175F897363D5659BEB54CA0}

SecureElement:
/send 80520000#(912015FB87F3577AC358E209EED7316C6EDB96981E04B4ED7E5DF597677B0B90BC2B921058EAE71460719CF332C9206863BD50F093473045022100C67639FA671CCCFC91B8A0CBFFFF5553EFBDC143933170C7129C48A42B45E3BE02204B12629A498DF0974E6B9DA4F915DC0C64D4C07E9C167E2FFA354B2888B9F79C)

tlv/tlv.go
MSalopek: Func was removed, so not relevant any more.

tlv/tlv.go
MSalopek: To continue the discussion, I actually removed `parsePrefix` function. The parsing is done in a for loop, similar to how it was done in the  previous TLV parser.

I also removed the things that made the doce look nice, because they were making the code more difficult to understand, but looked a lot prettier.

PR #168, opened 2022-07-19
Improve OpenSecureConnection to return error if card is uninitialized…

maevyn11:
… after selecting, instead of proceeding with doomed pairing attempt. 

simple improvement to robustness I ran into while smoke testing Douglas' testnet cards.


npmaile (approved):
very cash money

Issue #110, opened 2022-03-24
Update and Persist Session Names to Card


douglance:
Users want to be able to rename sessions from random hashes to something memorable. 

In order for the data to persist over time, updated names should be saved to the card itself. 

The frontend should receive these names when listing the sessions.

PR #169, opened 2022-07-19
Friendly name (#110)

zerodouglas:
Renames GetName to GetCardId. I added two new functions to the session interface so that the new resources to appropriately be called and added necessary changes to Swagger.

### DIFF
--- a/gui/api.go
+++ b/gui/api.go
@@ -67,7 +67,7 @@ func Server(port string, certFile string, keyFile string, mock bool) {
 		}
 		//will only be one
 		for _, sess := range session.t.ListSessions() {
-			cache[sess.GetName()] = &sessionCache{
+			cache[sess.GetCardId()] = &sessionCache{
 				phonons:        make(map[uint16]*model.Phonon),
 				cachePopulated: false,
 			}
@@ -78,7 +78,7 @@ func Server(port string, certFile string, keyFile string, mock bool) {
 			log.Error("unable to refresh card sessions during REST server startup: ", err)
 		}
 		for _, session := range sessions {
-			cache[session.GetName()] = &sessionCache{
+			cache[session.GetCardId()] = &sessionCache{
 				phonons:        make(map[uint16]*model.Phonon),
 				cachePopulated: false,
 			}
@@ -99,6 +99,8 @@ func Server(port string, certFile string, keyFile string, mock bool) {
 	r.HandleFunc("/cards/{sessionID}/init", session.init)
 	r.HandleFunc("/cards/{sessionID}/unlock", session.unlock)
 	r.HandleFunc("/cards/{sessionID}/pair", session.pair)
+	r.HandleFunc("/cards/{sessionID}/name", session.name)
+	r.HandleFunc("/cards/{sessionID}/name/create", session.createName)
 	// phonons
 	r.HandleFunc("/cards/{sessionID}/listPhonons", session.listPhonons)
 	r.HandleFunc("/cards/{sessionID}/phonon/{PhononIndex}/setDescriptor", session.setDescriptor)
@@ -238,7 +240,7 @@ func (apiSession apiSession) createPhonon(w http.ResponseWriter, r *http.Request
 		return
 	}
 
-	cache[sess.GetName()].phonons[index] = &model.Phonon{
+	cache[sess.GetCardId()].phonons[index] = &model.Phonon{
 		KeyIndex: index,
 		PubKey:   pubKey,
 	}
@@ -277,7 +279,7 @@ func (apiSession *apiSession) initDepositPhonons(w http.ResponseWriter, r *http.
 	}
 
 	for _, phonon := range phonons {
-		cache[sess.GetName()].phonons[phonon.KeyIndex] = phonon
+		cache[sess.GetCardId()].phonons[phonon.KeyIndex] = phonon
 	}
 
 	enc := json.NewEncoder(w)
@@ -414,7 +416,7 @@ func (apiSession apiSession) listSessions(w http.ResponseWriter, r *http.Request
 	for _, v := range sessions {
 		sessionStatuses = append(sessionStatuses,
 			&SessionStatus{
-				Name:           v.GetName(),
+				Name:           v.GetCardId(),
 				Initialized:    v.IsInitialized(),
 				TerminalPaired: v.IsPairedToTerminal(),
 				PinVerified:    v.IsUnlocked(),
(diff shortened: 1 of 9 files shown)


gui/api.go
```diff
@@ -99,6 +99,8 @@ func Server(port string, certFile string, keyFile string, mock bool) {
 	r.HandleFunc("/cards/{sessionID}/init", session.init)
 	r.HandleFunc("/cards/{sessionID}/unlock", session.unlock)
 	r.HandleFunc("/cards/{sessionID}/pair", session.pair)
+	r.HandleFunc("/cards/{sessionID}/name", session.name)
```
npmaile, line 102: just have the one endpoint of '/cards/{sessionID}/name'
use the verb to determine what to do. A GET should retrieve the name and a Post should set/overwrite the name

gui/api.go
```diff
@@ -587,6 +589,58 @@ func (apiSession apiSession) pair(w http.ResponseWriter, r *http.Request) {
...
+		Name string
+	}{}
+	err = json.Unmarshal(body, &nameReq)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+	}
+
+	err = sess.SetFriendlyName(nameReq.Name)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusUnprocessableEntity)
```
npmaile, line 614: just do a statusInternalServerError here though unprocessable entity may be more correct, but we would need to parse the error here to determine the precise issue before giving a more specific error response

gui/api.go
```diff
@@ -587,6 +589,58 @@ func (apiSession apiSession) pair(w http.ResponseWriter, r *http.Request) {
...
+func (apiSession apiSession) name(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	sess, err := apiSession.sessionFromMuxVars(vars)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusNotFound)
+		return
+	}
+	name, err := sess.GetFriendlyName()
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusUnprocessableEntity)
```
npmaile, line 628: same complaint as above

gui/api.go
```diff
@@ -587,6 +589,58 @@ func (apiSession apiSession) pair(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	if name == "" {
+		http.Error(w, "No friendly name found", http.StatusNotFound)
```
npmaile, line 632: after the change to cardID for the card's id, you can refer to the friendly name as "name"

npmaile (changes requested):
You've done precisely what I wanted. Good work. You just need to change a couple of things around the http semantics and error messaging

gui/api.go
```diff
@@ -587,6 +589,58 @@ func (apiSession apiSession) pair(w http.ResponseWriter, r *http.Request) {
 
 }
 
+func (apiSession apiSession) createName(w http.ResponseWriter, r *http.Request) {
```
maevyn11, line 592: I'd prefer setName or setFriendlyName for this function. create implies a new name is being created, but this call could also change an existing friendly name setting.

gui/api.go
```diff
@@ -587,6 +589,58 @@ func (apiSession apiSession) pair(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	err = sess.SetFriendlyName(nameReq.Name)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusUnprocessableEntity)
+		return
+	}
+}
+
+func (apiSession apiSession) name(w http.ResponseWriter, r *http.Request) {
```
maevyn11, line 619: Would prefer getName or getFriendlyName for this to make it clear from outside what it's doing.

zerodouglas: I suppose `getFriendlyName` is more clear about it's function.

gui/api.go
zerodouglas: I agree, I'll change to `setFriendlyName` to inline with `getFriendlyName`.

gui/api.go
zerodouglas: Sure, I'll get that changed to `sess.GetFriendName()` to `sess.GetName()` and `sess.SetName()`.

gui/swagger.yaml
```diff
@@ -199,6 +199,46 @@ paths:
       type: string
       description: sessionID of connected card
 
+  /cards/{sessionID}/name:
+    get:
+      tags:
+      - "sessions"
+      responses:
+        "500":
```
maevyn11, line 207: If you change the error code in the handler make sure this response code matches

douglance:
Testing this out with swagger, it seems that after renaming a card, then calling `/listSessions` the `phonon-client` returns each session's original name, rather than the name it has been renamed to. The frontend will have to make another request for each session, which isn't a big deal; but we have two things with the same name. This seems a bit confusing. 

I would like to propose having `/listSessions` return `Id` as the initial identifier and `Name` as a string of it's friendly name or an empty string if it isn't set (it could also just be absent if not set).

So the change would be from 
```json
[{"Name":"043bf3fe28725a70","Initialized":true,"TerminalPaired":true,"PinVerified":true}]
```
to:
```json
[{"Id":"043bf3fe28725a70","Name":"Friendly","Initialized":true,"TerminalPaired":true,"PinVerified":true}]
```
or:
```json
[{"Id":"043bf3fe28725a70","Name":"","Initialized":true,"TerminalPaired":true,"PinVerified":true}]
```

zerodouglas:
> Testing this out with swagger, it seems that after renaming a card, then calling `/listSessions` the `phonon-client` returns each session's original name, rather than the name it has been renamed to. The frontend will have to make another request for each session, which isn't a big deal; but we have two things with the same name. This seems a bit confusing.
> 
> I would like to propose having `/listSessions` return `Id` as the initial identifier and `Name` as a string of it's friendly name or an empty string if it isn't set (it could also just be absent if not set).
> 
> So the change would be from
> 
> ```json
> [{"Name":"043bf3fe28725a70","Initialized":true,"TerminalPaired":true,"PinVerified":true}]
> ```
> 
> to:
> 
> ```json
> [{"Id":"043bf3fe28725a70","Name":"Friendly","Initialized":true,"TerminalPaired":true,"PinVerified":true}]
> ```
> 
> or:
> 
> ```json
> [{"Id":"043bf3fe28725a70","Name":"","Initialized":true,"TerminalPaired":true,"PinVerified":true}]
> ```
> 
> This would also make more sense with the naming of `GetCardId()` (if I understand the changes correctly (I do not write Go)).

I think this makes sense and not a hard feature to implement. @npmaile @maevyn11 should we cut the function to only act as a POST in this case?

maevyn11:
Agree with @douglance 's suggestions as to renaming the fields passed to the frontend to disambiguate CardId and Name, passing both when available. 

Not sure what you mean by this? 

> should we cut the function to only act as a POST in this case?

Is the idea that the GET is unnecessary because the name will now be retrieved instead by /listSessions? 

I'm on board with that if so, better to have that functionality encapsulated in one place, but it does mean you'll need to ensure that the friendlyName will always have to be pulled into the session when it's available, so this call may need to be added to the session initialization sequence to ensure it's consistently cached in the client.

zerodouglas:
> Agree with @douglance 's suggestions as to renaming the fields passed to the frontend to disambiguate CardId and Name, passing both when available.
> 
> Not sure what you mean by this?
> 
> > should we cut the function to only act as a POST in this case?
> 
> Is the idea that the GET is unnecessary because the name will now be retrieved instead by /listSessions?
> 
> I'm on board with that if so, better to have that functionality encapsulated in one place, but it does mean you'll need to ensure that the friendlyName will always have to be pulled into the session when it's available, so this call may need to be added to the session initialization sequence to ensure it's consistently cached in the client.

Yes, I had removed the GET request endpoint as it's now retrieved via /listSessions. I just made the changes to support your suggestion. The call will specifically return it from the session sequence.

maevyn11:
All the code in the new changes looks good, but I think we should get the corresponding phonon-app changes committed to the submodule before merging, that way master will continue to function through the merge of the PR. https://github.com/GridPlus/phonon-app/pull/31 

Also conflicts with the other recent PRs need to be resolved before this one can merge.

maevyn11 (approved):
Tried it out and it works for me. Nicely done

npmaile:
fix merge conflicts and ill test again and approve

npmaile:
> fix merge conflicts and ill test again and approve

disregard. Force merging because I'll fix ci in a few minutes

PR #177, opened 2022-07-25
Friendly name support in repl

zerodouglas:


repl/repl.go
```diff
@@ -128,7 +134,11 @@ func Start() {
 func setActiveCard(c *ishell.Context, s *orchestrator.Session) {
 	activeCard = s
 	updatePrompt()
-	c.Printf("%v selected\n", activeCard.GetCardId())
+	if activeCard.GetName() != "" {
+		c.Printf("%v selected\n", activeCard.GetName())
+	} else {
+		c.Printf("%v selected\n", activeCard.GetCardId())
+	}
```
maevyn11, line 141: This logic is sort of being repeated a few times, so I think it would be best to wrap it up into a helper function and define the concept. 

```
func getDisplayName(activeCard *orchestrator.Session) string {
var name = activeCard.GetName()
if name != "" {
  return name 
} else {
  return activeCard.GetCardId()
}
  ```

Comment (maevyn11) [reply repl/repl.go:141]: Then this becomes: 
```
c.Printf("%v selected\n", activeCard.getDisplayName()

repl/repl.go
```diff
@@ -137,6 +147,10 @@ func updatePrompt() {
 		shell.SetPrompt(standardPrompt)
 	}
 	cardName := activeCard.GetCardId()
+	if activeCard.GetName() != "" {
+		cardName = activeCard.GetName()
+	}
+
```
maevyn11, line 153: displayName helper function would work here as well.

repl/repl.go
```diff
@@ -187,18 +201,69 @@ func listCards(c *ishell.Context) {
...
+						c.Printf("%v\n", s.GetCardId())
+					}
+				} else {
+					c.Printf("%v\n", s.GetName())
+				}
+
+			} else {
+				c.Printf("%v\n", s.GetCardId())
+			}
+		}
```
maevyn11, line 219: Think it might be simpler just to display both of these together in this list rather than adding the argument with multiple modes. The TODO should at least be resolved or removed before merging. 
```
cardID1 - Name1/Blank if missing
cardID2 - Name2/Blank if missing
```

repl/repl.go
```diff
@@ -187,18 +201,69 @@ func listCards(c *ishell.Context) {
...
+	err := activeCard.SetName(name)
+	if err != nil {
+		c.Printf("error setting name: %v", err)
+	} else {
+		updatePrompt()
+		c.Printf("name set to %v", name)
+	}
+}
+
+func setActiveCardName(c *ishell.Context, s *orchestrator.Session) {
```
maevyn11, line 244: Looks like this function isn't called anywhere and may be redundant with the above?

repl/repl.go
```diff
@@ -187,18 +201,69 @@ func listCards(c *ishell.Context) {
...
 func activateCard(c *ishell.Context) {
 	sessions := t.ListSessions()
 	var sessionNames []string
 	for _, session := range sessions {
-		sessionNames = append(sessionNames, session.GetCardId())
+		if session.GetName() == "" {
+			sessionNames = append(sessionNames, session.GetCardId())
+		} else {
+			sessionNames = append(sessionNames, session.GetName())
+		}
```
maevyn11, line 266: another good place for getDisplayName helper function.

maevyn11 (changes requested):
Looks pretty good overall, made a few comments. Can't really test it atm since my go env is in arrears.

repl/repl.go
```diff
@@ -130,26 +130,26 @@ func Start() {
 	shell.Run()
 }
 
+func getDisplayName(activeCard *orchestrator.Session) string {
+	var name = activeCard.GetName()
+	if name != "" {
+		return activeCard.GetName()
```
maevyn11, line 136: can be `return name` instead of calling the GetName() function again.

repl/repl.go
```diff
@@ -130,26 +130,26 @@ func Start() {
...
 //internal bookkeeping method to set a card to receive subsequent commands
 func setActiveCard(c *ishell.Context, s *orchestrator.Session) {
 	activeCard = s
 	updatePrompt()
-	if activeCard.GetName() != "" {
-		c.Printf("%v selected\n", activeCard.GetName())
-	} else {
-		c.Printf("%v selected\n", activeCard.GetCardId())
-	}
+	c.Printf("%v selected\n", getDisplayName(activeCard))
```
maevyn11, line 145: looks nice

repl/repl.go
```diff
@@ -159,7 +159,7 @@ func updatePrompt() {
 	} else {
 		status = ""
 	}
-	shell.SetPrompt(cardName + status + ">")
+	shell.SetPrompt(fmt.Sprintf("%v%v>", getDisplayName(activeCard), status))
```
maevyn11, line 162: love the fmt.Sprintf cleanup

repl/repl.go
```diff
@@ -203,16 +203,9 @@ func listCards(c *ishell.Context) {
...
-					if c.Args[0] == "ids" {
-						c.Printf("%v\n", s.GetCardId())
-					}
-				} else {
-					c.Printf("%v\n", s.GetName())
-				}
-
+			var name = s.GetName()
+			if name != "" {
+				c.Printf("%v - %v\n", s.GetCardId(), name)
```
maevyn11, line 208: looks good

maevyn11 (changes requested):
One tiny variable change needed and this is good to go.

repl/repl.go
zerodouglas: Looks like this slipped my eyes. Changes at https://github.com/GridPlus/phonon-client/pull/177/commits/3e99e8cbc32d1fa1ca3cc2b028d2145d8d7b64d1#diff-f3738eaa02e99234edda29447e81abd0c95e16bed33ee6a499379d8b34219373R136

maevyn11 (approved):
Looks like a merge conflict with the frontend has to be resolved otherwise this is good to go.

PR #179, opened 2022-07-25
fix all linter errors in go files

npmaile:
updates to each of the go files to make the gopls linter stop complaining. The fixes are including but not limited to the following:
1. lowercasing error messages
2. turning long if/else statements into switch case
3. assigning unused identifiers to the "_" sink
4. removing unnecessary sprint statements

### DIFF
--- a/card/apdu.go
+++ b/card/apdu.go
@@ -243,7 +243,7 @@ func NewCommandListPhonons(p1 byte, p2 byte, data []byte) *Command {
 			SW_WRONG_DATA + 3:           errors.New("unable to decode less than TLV"),
 			SW_WRONG_DATA + 4:           errors.New("unable to decode greater than TLV"),
 			SW_CONDITIONS_NOT_SATISFIED: ErrPINNotEntered,
-			SW_INCORRECT_P1P2:           errors.New("incorrect Parameters received"),
+			SW_INCORRECT_P1P2:           errors.New("incorrect parameters received"),
 		},
 	}
 }
@@ -329,7 +329,7 @@ func NewCommandReceivePhonons(phononTransferPacket []byte) *Command {
 		PossibleErrs: CmdErrTable{
 			SW_CONDITIONS_NOT_SATISFIED: errors.New("phonon recipt conditions not met"),
 			SW_FILE_FULL:                errors.New("maximum number of phonons exceeded"),
-			SW_WRONG_DATA:               errors.New("unable to decode Phonon key list TLV"),
+			SW_WRONG_DATA:               errors.New("unable to decode phonon key list TLV"),
 		},
 	}
 }
@@ -415,7 +415,7 @@ func NewCommandCardPair2(data []byte) *Command {
 			SW_CONDITIONS_NOT_SATISFIED: ErrPINNotEntered,
 			SW_WRONG_DATA:               errors.New("unable to read salt"),
 			SW_WRONG_DATA + 1:           errors.New("unable to read AES TLV"),
-			SW_WRONG_DATA + 2:           errors.New("unable to read Signature TLV"),
+			SW_WRONG_DATA + 2:           errors.New("unable to read signature TLV"),
 		},
 	}
 }
@@ -432,7 +432,7 @@ func NewCommandFinalizeCardPair(data []byte) *Command {
 		PossibleErrs: CmdErrTable{
 			// No idea how you can get this far without validating a pin
 			SW_CONDITIONS_NOT_SATISFIED:      ErrPINNotEntered,
-			SW_WRONG_DATA:                    errors.New("unable to read Receiver signature TLV"),
+			SW_WRONG_DATA:                    errors.New("unable to read receiver signature TLV"),
 			SW_SECURITY_STATUS_NOT_SATISFIED: errors.New("unable to verify signature"),
 		},
 	}
@@ -513,7 +513,7 @@ func NewCommandMutualAuthenticate(data []byte) *Command {
 		ApduCmd: keycard.NewCommandMutuallyAuthenticate(data),
 		PossibleErrs: CmdErrTable{
 			SW_CONDITIONS_NOT_SATISFIED:      errors.New("authentication key not initialized"),
-			SW_LOGICAL_CHANNEL_NOT_SUPPORTED: errors.New("already Mutually Authenticated"),
+			SW_LOGICAL_CHANNEL_NOT_SUPPORTED: errors.New("already mutually authenticated"),
 			SW_SECURITY_STATUS_NOT_SATISFIED: errors.New("secret length invalid"),
 		},
 	}
--- a/card/apdu.go
+++ b/card/apdu.go
@@ -238,12 +238,12 @@ func NewCommandListPhonons(p1 byte, p2 byte, data []byte) *Command {
 		),
 		PossibleErrs: CmdErrTable{
 			SW_WRONG_DATA:               errors.New("no remaining phonons to list"),
-			SW_WRONG_DATA + 1:           errors.New("Unable to decode phonon filter TLV"),
+			SW_WRONG_DATA + 1:           errors.New("unable to decode phonon filter TLV"),
 			SW_WRONG_DATA + 2:           errors.New("unable to decode phonon currency TLV"),
-			SW_WRONG_DATA + 3:           errors.New("Unable to decode less than TLV"),
-			SW_WRONG_DATA + 4:           errors.New("Unable to decode greater than TLV"),
+			SW_WRONG_DATA + 3:           errors.New("unable to decode less than TLV"),
+			SW_WRONG_DATA + 4:           errors.New("unable to decode greater than TLV"),
 			SW_CONDITIONS_NOT_SATISFIED: ErrPINNotEntered,
-			SW_INCORRECT_P1P2:           errors.New("Incorrect Parameters received"),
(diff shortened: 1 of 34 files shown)


card/apdu.go
```diff
@@ -238,12 +238,12 @@ func NewCommandListPhonons(p1 byte, p2 byte, data []byte) *Command {
...
-			SW_WRONG_DATA + 1:           errors.New("Unable to decode phonon filter TLV"),
+			SW_WRONG_DATA + 1:           errors.New("unable to decode phonon filter TLV"),
 			SW_WRONG_DATA + 2:           errors.New("unable to decode phonon currency TLV"),
-			SW_WRONG_DATA + 3:           errors.New("Unable to decode less than TLV"),
-			SW_WRONG_DATA + 4:           errors.New("Unable to decode greater than TLV"),
+			SW_WRONG_DATA + 3:           errors.New("unable to decode less than TLV"),
+			SW_WRONG_DATA + 4:           errors.New("unable to decode greater than TLV"),
 			SW_CONDITIONS_NOT_SATISFIED: ErrPINNotEntered,
-			SW_INCORRECT_P1P2:           errors.New("Incorrect Parameters received"),
+			SW_INCORRECT_P1P2:           errors.New("incorrect Parameters received"),
```
maevyn11, line 246: "Parameters should be lowercased too.

card/apdu.go
```diff
@@ -327,9 +327,9 @@ func NewCommandReceivePhonons(phononTransferPacket []byte) *Command {
 			phononTransferPacket,
 		),
 		PossibleErrs: CmdErrTable{
-			SW_CONDITIONS_NOT_SATISFIED: errors.New("Phonon recipt conditions not met"),
-			SW_FILE_FULL:                errors.New("Maximum number of phonons exceeded"),
-			SW_WRONG_DATA:               errors.New("Unable to decode Phonon key list TLV"),
+			SW_CONDITIONS_NOT_SATISFIED: errors.New("phonon recipt conditions not met"),
+			SW_FILE_FULL:                errors.New("maximum number of phonons exceeded"),
+			SW_WRONG_DATA:               errors.New("unable to decode Phonon key list TLV"),
```
maevyn11, line 332: "Phonon" should be lowercased

card/apdu.go
```diff
@@ -413,9 +413,9 @@ func NewCommandCardPair2(data []byte) *Command {
 		),
 		PossibleErrs: CmdErrTable{
 			SW_CONDITIONS_NOT_SATISFIED: ErrPINNotEntered,
-			SW_WRONG_DATA:               errors.New("Unable to read salt"),
-			SW_WRONG_DATA + 1:           errors.New("Unable to read AES TLV"),
-			SW_WRONG_DATA + 2:           errors.New("Unable to read Signature TLV"),
+			SW_WRONG_DATA:               errors.New("unable to read salt"),
+			SW_WRONG_DATA + 1:           errors.New("unable to read AES TLV"),
+			SW_WRONG_DATA + 2:           errors.New("unable to read Signature TLV"),
```
maevyn11, line 418: "Signature" caps

card/apdu.go
```diff
@@ -432,8 +432,8 @@ func NewCommandFinalizeCardPair(data []byte) *Command {
 		PossibleErrs: CmdErrTable{
 			// No idea how you can get this far without validating a pin
 			SW_CONDITIONS_NOT_SATISFIED:      ErrPINNotEntered,
-			SW_WRONG_DATA:                    errors.New("Unable to read Receiver signature TLV"),
-			SW_SECURITY_STATUS_NOT_SATISFIED: errors.New("Unable to verify signature"),
+			SW_WRONG_DATA:                    errors.New("unable to read Receiver signature TLV"),
```
maevyn11, line 435: "Receiver" caps

card/apdu.go
```diff
@@ -512,9 +512,9 @@ func NewCommandMutualAuthenticate(data []byte) *Command {
 	return &Command{
 		ApduCmd: keycard.NewCommandMutuallyAuthenticate(data),
 		PossibleErrs: CmdErrTable{
-			SW_CONDITIONS_NOT_SATISFIED:      errors.New("Authentication key not initialized"),
-			SW_LOGICAL_CHANNEL_NOT_SUPPORTED: errors.New("Already Mutually Authenticated"),
-			SW_SECURITY_STATUS_NOT_SATISFIED: errors.New("Secret length invalid"),
+			SW_CONDITIONS_NOT_SATISFIED:      errors.New("authentication key not initialized"),
+			SW_LOGICAL_CHANNEL_NOT_SUPPORTED: errors.New("already Mutually Authenticated"),
```
maevyn11, line 516: "Mutually Authenticated" caps

card/phononCommandSet_test.go
```diff
@@ -95,12 +95,12 @@ func TestCreateSetAndListPhonons(t *testing.T) {
...
-		{model.Bitcoin, model.Denomination{99, 6}},
-		{model.Ethereum, model.Denomination{1, 18}},
-		{model.Ethereum, model.Denomination{9, 18}},
-		{model.Ethereum, model.Denomination{1, 0}},
+		{model.Bitcoin, model.Denomination{Base: 1, Exponent: 0}},
+		{model.Bitcoin, model.Denomination{Base: 1, Exponent: 8}},
+		{model.Bitcoin, model.Denomination{Base: 99, Exponent: 6}},
+		{model.Ethereum, model.Denomination{Base: 1, Exponent: 18}},
+		{model.Ethereum, model.Denomination{Base: 9, Exponent: 18}},
+		{model.Ethereum, model.Denomination{Base: 1, Exponent: 0}},
```
maevyn11, line 103: Nice

card/phononCommandSet_test.go
```diff
@@ -274,7 +274,8 @@ func TestFillPhononTable(t *testing.T) {
 	maxPhononCount := 256
 	var createdIndices []uint16
 	for i := 0; i < maxPhononCount-initialCount; i++ {
-		keyIndex, _, err := cs.CreatePhonon(model.Secp256k1)
+		var keyIndex uint16
+		keyIndex, _, err = cs.CreatePhonon(model.Secp256k1)
```
maevyn11, line 278: why this change? Isn't this equivalent?

maevyn11: old way seems cleaner to me.

cert/cert.go
```diff
@@ -140,8 +140,8 @@ func SignWithDemoKey(cert []byte) ([]byte, error) {
 
 	// print("signing with demo key")
 	key.D = new(big.Int).SetBytes(demoKey)
-	key.PublicKey.Curve = secp256k1.S256()
-	key.PublicKey.X, key.PublicKey.Y = key.PublicKey.Curve.ScalarBaseMult(key.D.Bytes())
+	key.Curve = secp256k1.S256()
+	key.X, key.Y = key.ScalarBaseMult(key.D.Bytes())
```
maevyn11, line 144: We sure this is equivalent? I guess we're accessing the embedded fields here, but doing it explicitly on the public key seems more clear, whereas now I'm not sure if Curve is at the top level, or in the private key, or the public key or some combination.

cert/cert.go
```diff
@@ -198,7 +198,7 @@ func ParseRawCardCertificate(cardCertificateRaw []byte) (cert CardCertificate, e
 
 	if cert.Permissions.certLen == 0 || cert.Permissions.permLen == 0 {
 		log.Debugf("invalid certificate found: % X", cardCertificateRaw)
-		return CardCertificate{}, errors.New("Card Certificate Invalid")
+		return CardCertificate{}, errors.New("card Certificate Invalid")
```
maevyn11, line 201: caps on the last two words

util/crypto.go
```diff
@@ -33,24 +33,26 @@ func ParseECCPubKey(rawPubKey []byte) (pubKey *ecdsa.PublicKey, err error) {
...
 		pubKey, err = ethcrypto.DecompressPubkey(rawPubKey)
 		if err != nil {
 			log.Error("could not unmarshal compressed ecdsa pub key from raw: ", err)
 			return nil, err
 		}
-	} else {
+	default:
 		log.Debugf("could not detect ECC pubkey format from key: % X\n", rawPubKey)
 		return nil, ErrInvalidECCPubKeyFormat
+
```
maevyn11, line 55: This is a nice change but it scares me a bit since it touches crypto parsing which could be hard to debug from outside. Is there a unit test to regression test this still works the same?

maevyn11 (changes requested):
Thanks for doing this, few minor comments and it should be good to go.

card/phononCommandSet_test.go
npmaile: it was mad because err was shaddowed. This is more explicit. 
I really just don't like the linter complaining

cert/cert.go
npmaile: yes. the private key (according to the go implementation of the ecdsa library) consists of an embedded public key and a D field.

npmaile: https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/crypto/ecdsa/ecdsa.go;drc=8384fe86a5b7f579a50c7ad423d4dd4eb2d1f117;l=80

util/crypto.go
npmaile: there is a unit test, and it does pass.

maevyn11 (approved):
Looks great

Issue #178, opened 2022-07-25
Fetch Friendly Name at Session Start


maevyn11:
It appears the issue described in https://github.com/GridPlus/phonon-app/issues/32 is at least partially the fault of the client. 

As stated in this phonon-app issue: https://github.com/GridPlus/phonon-app/issues/32, the frontend's initial call to /listSessions fetches an empty `Name` field, even when the client has one set. I suspect the issue arises because GetFriendlyName is not being called as part of the NewSession initialization sequence, and the `Name` field is therefore only set after the SetFriendlyName flow is initiated, whereas it should also be set on startup.

Issue #170, opened 2022-07-20
Phonons show "Native" after Creation Instead of correct ChainID


maevyn11:
To reproduce, create a phonon on Rinkeby ETH, the icon next to it shows as "Native" instead of Rinkeby. If you restart the server and unlock the card again, the correct icon is shown to the left.

Think this actually happens in the client when phonons are being returned right after deposit. Leading suspect according to Nate is the caching layer within the REST API server. 

Moved over from this phonon-app issue as the problem actually appears to be in the client. https://github.com/GridPlus/phonon-app/issues/24

PR #180, opened 2022-07-25
Move cache

zerodouglas:
Migrated cache out of the API and into the orchestrator. Resolves numerous issues such as #170 and #178. I did test the functionality, you can indeed create a phonon with the proper chain id, pair cards and send to that card id as well as redeeming.

### DIFF
--- a/gui/api.go
+++ b/gui/api.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"html/template"
+	"io"
 	"io/fs"
 	"io/ioutil"
 	"math/big"
@@ -91,6 +92,7 @@ func Server(port string, certFile string, keyFile string, mock bool) {
 	r.HandleFunc("/cards/{sessionID}/connect", session.ConnectRemote)
 	r.HandleFunc("/cards/{sessionID}/connectionStatus", session.RemoteConnectionStatus)
 	r.HandleFunc("/cards/{sessionID}/connectLocal", session.ConnectLocal)
+	r.HandleFunc("/checkDenomination", verifyDenomination)
 	// api docs
 	r.PathPrefix("/swagger/").Handler(http.StripPrefix("/", http.FileServer(http.FS(swagger))))
 	r.HandleFunc("/swagger.json", serveAPIFunc(port))
@@ -122,6 +124,27 @@ func Server(port string, certFile string, keyFile string, mock bool) {
 	}()
 	browser.OpenURL("http://localhost:" + port + "/")
 	systray.Run(onReady, onExit)
+}
+
+func verifyDenomination(w http.ResponseWriter, r *http.Request) {
+	tocheckBytes, err := io.ReadAll(r.Body)
+	if err != nil {
+		http.Error(w, "unable to parse request", http.StatusBadRequest)
+		return
+	}
+	toCheckString := string(tocheckBytes)
+	toCheck := new(big.Int)
+	toCheck, ok := toCheck.SetString(toCheckString, 10)
+	if !ok {
+		http.Error(w, "unable to coerce request string to integer", http.StatusBadRequest)
+		return
+	}
+	_, err = model.NewDenomination(toCheck)
+	if err != nil {
+		http.Error(w, "cannot make phonon denomination from value: "+err.Error(), http.StatusBadRequest)
+		return
+	}
+	// return 200 if it works
 }
 
 func logsink(w http.ResponseWriter, r *http.Request) {
@@ -220,8 +243,8 @@ func (apiSession apiSession) createPhonon(w http.ResponseWriter, r *http.Request
 
 	enc := json.NewEncoder(w)
 	enc.Encode(struct {
-		Index  uint16 `json:"index"`
-		PubKey string `json:"pubkey"`
+		Index  model.PhononKeyIndex `json:"index"`
+		PubKey string               `json:"pubkey"`
 	}{Index: index,
 		PubKey: pubKey.String()})
 }
@@ -664,11 +687,11 @@ func (apiSession apiSession) setDescriptor(w http.ResponseWriter, r *http.Reques
 	}
 
 	p := &model.Phonon{
-		KeyIndex:     uint16(index),
+		KeyIndex:     model.PhononKeyIndex(index),
 		Denomination: den,
 		CurrencyType: model.CurrencyType(inputs.CurrencyType),
 	}
-	p.KeyIndex = uint16(index)
+	p.KeyIndex = model.PhononKeyIndex(index)
 	err = sess.SetDescriptor(p)
 	if err != nil {
 		http.Error(w, "Unable to set descriptor", http.StatusBadRequest)
@@ -690,7 +713,7 @@ func (apiSession apiSession) send(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	json.Unmarshal(bodyBytes, &inputs)
-	toSend := []uint16{}
+	toSend := []model.PhononKeyIndex{}
 	for _, phonon2send := range inputs {
 		toSend = append(toSend, phonon2send.KeyIndex)
 	}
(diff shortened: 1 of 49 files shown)


npmaile (changes requested):
s.friendlyName is already the cached friendly name. I forgot to remove cachedFriendlyName from the struct after I erroneusly added it not realizing the existing friendly name field was already a cache.

orchestrator/session.go
```diff
@@ -122,21 +133,33 @@ func (s *Session) GetCardId() string {
 	}
 }
 
-func (s *Session) GetName() string {
-	return s.friendlyName
+func (s *Session) GetName() (string, error) {
+	if s.friendlyName != "" {
+		return s.cachedFriendlyName, nil
```
npmaile, line 138: need to change this reference to friendlyName

orchestrator/session.go
```diff
@@ -122,21 +133,33 @@ func (s *Session) GetCardId() string {
...
+	} else {
+		var err error
+		s.friendlyName, err = s.cs.GetFriendlyName()
+		if err != nil {
+			return "", err
+		}
+
+		s.cachedFriendlyName = s.friendlyName
```
npmaile, line 146: this isn't necessary

orchestrator/session.go
zerodouglas: > s.friendlyName is already the cached friendly name. I forgot to remove cachedFriendlyName from the struct after I erroneusly added it not realizing the existing friendly name field was already a cache.

Makes more sense, I wasn't sure of the use for cachedFriendlyName

orchestrator/session.go
```diff
@@ -122,21 +132,31 @@ func (s *Session) GetCardId() string {
...
+	if s.friendlyName != "" {
+		return s.friendlyName, nil
+	} else {
+		var err error
+		s.friendlyName, err = s.cs.GetFriendlyName()
+		if err != nil {
+			return "", err
+		}
+	}
+	return s.friendlyName, nil
```
maevyn11, line 145: This is to fix the missing friendlyName on startup right? Pattern becomes the same as GetCardID() where it will run a command to check if the name isn't already cached?

orchestrator/session.go
```diff
@@ -122,21 +132,31 @@ func (s *Session) GetCardId() string {
...
 		return ErrNameCannotBeEmpty
 	}
-
-	s.friendlyName = name
-	return s.cs.SetFriendlyName(name)
+	err := s.cs.SetFriendlyName(name)
+	if err == nil {
+		s.friendlyName = name
+	}
+	return nil
```
maevyn11, line 159: This is a nice fix to ensure that the name isn't cached if there is an error talking to the card but I think it breaks error handling for SetName(), since this would now always return nil, even if s.cs.SetFriendlyName returns an error. 

changing line 159 to the following would fix. 
```
return err
```

orchestrator/session.go
```diff
@@ -380,6 +443,8 @@ func (s *Session) ReceivePhonons(phononTransferPacket []byte) error {
 	if err != nil {
 		return err
 	}
+	//invalidate the cache now that new phonons have been received
+	s.cachePopulated = false
```
maevyn11, line 447: cool, was wondering where this happens and this spot makes a lot of sense.

orchestrator/session.go
```diff
@@ -655,3 +720,44 @@ func (s *Session) RedeemPhonon(p *model.Phonon, redeemAddress string) (transacti
 }
 
 //TODO: retry and track progress automatically.
+
+/*addPubKeyToCache adds the pubkey to an already cached phonon. This is done differently from the addInfoToCache for reasons.*/
```
maevyn11, line 724: what reasons?

orchestrator/session.go
```diff
@@ -254,17 +285,38 @@ func (s *Session) SetDescriptor(p *model.Phonon) error {
...
 	defer s.ElementUsageMtex.Unlock()
 
-	return s.cs.ListPhonons(currencyType, lessThanValue, greaterThanValue, false)
+	phonons, err := s.cs.ListPhonons(currencyType, lessThanValue, greaterThanValue, false)
+	// add listed phonons to the cache
+	for _, phonon := range phonons {
+		s.addInfoToCache(phonon)
+	}
+
+	if currencyType == 0 && lessThanValue == 0 && greaterThanValue == 0 {
```
maevyn11, line 315: good check, this comment is for a larger discussion, but now that we have this caching layer to improve performance, I think we may want to ditch the on card filtering entirely for Beta. It's unimplemented anyway, we haven't had a use for it, and it seems to mostly serve to add complexity like this this statement at this point.

orchestrator/session.go
```diff
@@ -655,3 +720,44 @@ func (s *Session) RedeemPhonon(p *model.Phonon, redeemAddress string) (transacti
...
+	cached, ok := s.cache[p.KeyIndex]
+	// if it didn't already exist, create it
+	if !ok {
+		s.cache[p.KeyIndex] = cachedPhonon{
+			p:          p,
+			infoCached: true,
+		}
+		// otherwise, add the new descriptor things
+	} else {
+		cachedPubKey := cached.p.PubKey
```
maevyn11, line 754: Is there any risk that accessing this value here will nil panic, either on p or on pubkey? 

Little confused about the handling of the PubKey value within this function when there is already addPubKeyToCache above? Feel like I'm missing why setting this would be relevant here.

gui/api.go
```diff
@@ -632,29 +607,19 @@ func (apiSession apiSession) listPhonons(w http.ResponseWriter, r *http.Request)
...
-		phonons, err = sess.ListPhonons(0, 0, 0)
+	var phonons []*model.Phonon
+	phonons, err = sess.ListPhonons(0, 0, 0)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	for _, p := range phonons {
+		p.PubKey, err = sess.GetPhononPubKey(p.KeyIndex, p.CurveType)
```
maevyn11, line 618: Not for this PR, but it seems to me that always call GetPhononPubKey along with ListPhonons. Might simplify things to merge these under one higher level Session call or actually merge these into one command on the card for Beta

orchestrator/session.go
```diff
@@ -37,12 +37,21 @@ type Session struct {
...
 	logger                *log.Entry
 	chainSrv              chain.ChainService
+	cache                 map[uint16]cachedPhonon
+	// cachePopulated indicates if all of the phonons present on the card have been cached. This is currently only set when listphonons is called with the values to list all phonons on the card.
+	cachePopulated bool
+}
+
+type cachedPhonon struct {
+	pubkeyCached bool
+	infoCached   bool
```
maevyn11, line 47: I don't see pubkeyCached or infoCached being checked anywhere. Only a check on the whole cache with cachePopulated. 

Plans for future expansion with these or is there another reason why they're included?

maevyn11 (commented):
Just asked for clarification on a couple elements of the design. Can probably go in as is once we clear those up.

orchestrator/session.go
npmaile: exactly

orchestrator/session.go
npmaile: :+1:

orchestrator/session.go
npmaile: it will set it to nil if it's already nil. the two functions exist so that if we have the pubkey cached and not the rest of the info, we cache the rest of the info without changing the pubkey or asking for it again, and if we have the info cached, but not the pubkey, we can cache the pubkey without changing everything else.

gui/api.go
npmaile: that would clear up the above confusion with the cachepubkey vs cachephonon stuff

orchestrator/session.go
maevyn11: should explain the reasons in the comment or remove the mysterious reference. This comment will likely be frustrating to look at in 6 months when we've all forgot the reasons

maevyn11 (approved):
Code's good to go but left one suggestion to improve a comment.

maevyn11:
Hey I checked out the latest version of this branch to test and I the bugs https://github.com/GridPlus/phonon-client/issues/170 and https://github.com/GridPlus/phonon-client/issues/178 still seem to be present for me. Can someone else pull and let me know if you're seeing this or not?

npmaile:
crap, actions is failing now

npmaile (approved):

PR #182, opened 2022-07-26
add verify denomination endpoint

npmaile:
Add endpoint to verify denominations
returns 400 on bad request or bad number (with explanation)
returns 200 when it's good

### DIFF
--- a/gui/api.go
+++ b/gui/api.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"html/template"
+	"io"
 	"io/fs"
 	"io/ioutil"
 	"math/big"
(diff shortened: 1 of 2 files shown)


gui/api.go
```diff
@@ -145,6 +147,27 @@ func Server(port string, certFile string, keyFile string, mock bool) {
...
+		http.Error(w, "unable to parse request", http.StatusBadRequest)
+		return
+	}
+	toCheckString := string(tocheckBytes)
+	toCheck := new(big.Int)
+	toCheck, ok := toCheck.SetString(toCheckString, 10)
+	if !ok {
+		http.Error(w, "unable to coerce request string to integer", http.StatusBadRequest)
+		return
+	}
```
maevyn11, line 162: does this conversion from byte to string to big Int handle decimals like ".000001"?

npmaile: no

maevyn11: discussed out of band that this limitation is fine for now.

maevyn11 (approved):

PR #194, opened 2022-08-02
Removes mainnet and adds Goerli

zerodouglas:
This will need further implementation on the phonon-app from @douglance. However, it temporarily removes `mainnet` and introduces another ETH chain, Goerli.


zerodouglas:
Tested with submodule update. Create and Redeem works as intended.

maevyn11 (approved):

PR #195, opened 2022-08-03
Startup window

npmaile:
Window that opens on startup from desktop or optionally when using the command line option to allow for setting the telemetry key


cmd/graphicalConfigure.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/GridPlus/phonon-client/config"
+	"github.com/spf13/cobra"
+)
+
+// graphicalConfigureCmd represents the graphicalConfigure command
+var graphicalConfigureCmd = &cobra.Command{
+	Use:   "graphicalConfigure",
+	Short: "A brief description of your command",
+	Long: `A longer description that spans multiple lines and likely contains examples
+and usage of using your command. For example:
```
maevyn11, line 17: needs a description so CLI users know what they're looking at.

config/config.go
```diff
@@ -31,22 +32,27 @@ func DefaultConfig() Config {
 }
 
 func SetDefaultConfig() {
-	viper.SetDefault("AppletCACert", cert.PhononAlphaCAPubKey)
+	viper.SetDefault("AppletCACert", string(cert.PhononAlphaCAPubKey))
```
maevyn11, line 35: why is this changed to string? Isn't AppletCACert asking for a []byte?

config/graphical.go
```diff
@@ -0,0 +1,101 @@
...
+	"fyne.io/fyne/v2/widget"
+	"github.com/spf13/viper"
+)
+
+var loggingTestURL = "https://logs.phonon.network/testKey"
+
+func GraphicalConfiguration() {
+	a := app.New()
+	w := a.NewWindow("Configure Phonon Application")
+	welcome := widget.NewLabel("Welcome to Phonon configuration. Please paste your alpha logging key here")
```
maevyn11, line 24: key is referred to as "Telemetry Key" below, I would stick with the same name in all the user facing text to avoid confusion. Suggest "Please paste your telemetry key here"

config/graphical.go
```diff
@@ -0,0 +1,101 @@
...
+			return err
+		}
+	case "windows":
+		configPath = homedir + "\\.phonon\\"
+		os.MkdirAll(configPath, 0700)
+	default:
+		return fmt.Errorf("unable to set configuration path for %s", runtime.GOOS)
+	}
+	return viper.WriteConfigAs(configPath + "phonon.yml")
+}
```
maevyn11, line 75: I think this function belongs in the Config package where it can be reused. Also LoadConfig() and this should derive the homedir in the same way so they can't get out of sync. Suggest factoring out the homedir detection and directory creation into one or two other functions called by SaveConfig and LoadConfig

release/MacOS/Phonon.app/Contents/MacOS/phonon-launcher
```diff
@@ -1,10 +1,21 @@
...
 then
+	echo "a"
+	if ! [ -f "$HOME/.phonon/phonon.yml" ] ; then
+		echo "b"
+		"${0%/*}/phonon_arm64" graphicalConfigure
+		echo "c"
+	fi
+	echo "d"
 	exec "${0%/*}/phonon_arm64" webUI -p 8080
+	echo "e"
```
maevyn11, line 12: assume these echo statements are for debugging and should be removed before merge?

release/MacOS/Phonon.app/Contents/MacOS/phonon-launcher
```diff
@@ -1,10 +1,21 @@
 #!/usr/bin/env bash
 if [ `arch` == "arm64" ] 
 then
+	echo "a"
+	if ! [ -f "$HOME/.phonon/phonon.yml" ] ; then
```
maevyn11, line 5: Can this check the telemetry key test endpoint instead of checking for the config file like we discussed? 

Otherwise this would have to account for the alternate config files supported within Viper's config setup.

maevyn11 (changes requested):
Left a review on the code, going to run through the configurator myself and give some design feedback.

config/graphical.go
npmaile: it is in the config package

release/MacOS/Phonon.app/Contents/MacOS/phonon-launcher
npmaile: correct

config/graphical.go
```diff
@@ -0,0 +1,101 @@
...
+	welcome := widget.NewLabel("Welcome to Phonon configuration. Please paste your alpha logging key here")
+	keyBox := widget.NewEntry()
+	w.SetContent(container.NewVBox(
+		welcome,
+		keyBox,
+		widget.NewButton("Test Logging Key", func() {
+			cleaned := strings.Trim(keyBox.Text, `\ `)
+			err := CheckTelemetryKey(cleaned)
+			if err != nil {
+				welcome.SetText(fmt.Sprintf("Telemetry key validation failed: %s", err.Error()))
```
maevyn11, line 33: I don't seem to be able to get this error to show

config/graphical.go
```diff
@@ -0,0 +1,101 @@
...
+			}
```
maevyn11, line 34: Should be some confirmation text for the user when the logging key test succeeds.

release/MacOS/Phonon.app/Contents/MacOS/phonon-launcher
npmaile: not really. I'm assuming graphical users will not want to use an alternative config path, and if the check fails here, the application will fail to launch. the endpoint check will happen when it's being configured and in the app. The launcher really shouldn't be making calls to the network, and there's not really any way to notify the user of what's going on if it fails

config/graphical.go
maevyn11: right...didn't realize that at the time. Guess I expected it to be in config.go because this doesn't specifically pertain to the graphical frontend configurator.

config/graphical.go
maevyn11: nvm. I can get it to show, I just didn't realize that lack of output meant it was succeeding.

config/graphical.go
```diff
@@ -0,0 +1,101 @@
...
+func GraphicalConfiguration() {
+	a := app.New()
+	w := a.NewWindow("Configure Phonon Application")
+	welcome := widget.NewLabel("Welcome to Phonon configuration. Please paste your alpha logging key here")
+	keyBox := widget.NewEntry()
+	w.SetContent(container.NewVBox(
+		welcome,
+		keyBox,
+		widget.NewButton("Test Logging Key", func() {
+			cleaned := strings.Trim(keyBox.Text, `\ `)
```
maevyn11, line 30: If you open the graphicalConfigure dialog and you already have a key configured in the config file, but you don't paste it into the box, clicking this test returns an error for "no auth given." 

Ideally if a config key is already present, the test button will use that one. Maybe if the logging key input box is empty we can look in the existing config for a key to check, and return the check results based on that? 

In fact, we could drop the "Save Configuration" button, and only save the config when there has been a confirmed sucessful test of the logging key. I'm concerned some users will hit test but then skip hitting "Save" and not realize no config file has been written.

maevyn11 (changes requested):

release/MacOS/Phonon.app/Contents/MacOS/phonon-launcher
maevyn11: right makes sense. Yea network call here is a bad idea. 

This could cause problems for users who use both the GUI and the CLI and have an alternate config path, but we could fix that by replicating alternate config files into the "standard config location" Not a big deal for now as that will be an unlikely scenario, and we can cross that bridge when we actually encounter the problem.

maevyn11:
UI looks pretty good design wise. Idk what options exist in this dialog library, but if it could use the phonon brand colors to match more or less the frontend colors that'd be slick. 

Only other suggestion would be to drop the "Save Configuration". button and combine that workflow into one step with a successful test also saving the config, like I suggested in another comment.

config/config.go
```diff
@@ -82,3 +89,36 @@ func LoadConfig() (config Config, err error) {
...
+	configPath, err := DefaultConfigPath()
+	if err != nil {
+		return err
+	}
+	err = os.MkdirAll(configPath, 0700)
+	if err != nil {
+		return err
+	}
+	return viper.WriteConfigAs(configPath + "phonon.yml")
+}
```
maevyn11, line 124: Something in this config save is writing the appletcacert key as a decimal list that looks like this: 
```
appletcacert:
- 4
- 114
- 213
- 140
- 30
- 196
- 143
...
```
This should really be changed so we're just saving the "dev" or "alpha" config key and the appletcacert variable is not written to the config file.

maevyn11 (changes requested):

npmaile:
tested and working on windows and linux

config/config.go
npmaile: fixed

maevyn11 (approved):
Looks good!

PR #198, opened 2022-08-04
Added EVM tokens and a gas fee check

zerodouglas:
Adds Binance, Avalanche, and Polygon. It's to note that they will fail if the gas fee is too low for redemption, for this I added a small check to display to the gui exactly when this fails for the legacy transaction. I did test each and every and they do work as intended. Some of these tokens offers multiple RPC nodes such as Binance, however, I just stuck with the first one they provided.

### DIFF
--- a/chain/ethChainService.go
+++ b/chain/ethChainService.go
@@ -18,7 +18,7 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-//Composite interface supporting all needed EVM RPC calls
+// Composite interface supporting all needed EVM RPC calls
 type EthChainInterface interface {
 	bind.ContractTransactor
 	ethereum.ChainStateReader
@@ -38,7 +38,7 @@ func NewEthChainService() (*EthChainService, error) {
 	return ethchainSrv, nil
 }
 
-//Derives an ETH address from a phonon's ECDSA Public Key
+// Derives an ETH address from a phonon's ECDSA Public Key
 func (eth *EthChainService) DeriveAddress(p *model.Phonon) (address string, err error) {
 	eccPubKey, err := model.PhononPubKeyToECDSA(p.PubKey)
 	if err != nil {
@@ -83,8 +83,8 @@ func (eth *EthChainService) RedeemPhonon(p *model.Phonon, privKey *ecdsa.Private
 	return tx.Hash().String(), nil
 }
 
-//ReconcileRedeemData validates the input data to ensure it contains all that's needed for a successful redemption.
-//It will update the phonon data structure with a derived address if necessary
+// ReconcileRedeemData validates the input data to ensure it contains all that's needed for a successful redemption.
+// It will update the phonon data structure with a derived address if necessary
 func (eth *EthChainService) ValidateRedeemData(p *model.Phonon, privKey *ecdsa.PrivateKey, redeemAddress string) (err error) {
 	eccPubKey, err := model.PhononPubKeyToECDSA(p.PubKey)
 	if err != nil {
@@ -117,7 +117,7 @@ func (eth *EthChainService) ValidateRedeemData(p *model.Phonon, privKey *ecdsa.P
 	return nil
 }
 
-//dialRPCNode establishes a connection to the proper RPC node based on the chainID
+// dialRPCNode establishes a connection to the proper RPC node based on the chainID
 func (eth *EthChainService) dialRPCNode(chainID int) (err error) {
 	log.Debugf("ethChainID: %v, chainID: %v\n", eth.clChainID, chainID)
 	var RPCEndpoint string
@@ -142,6 +142,8 @@ func (eth *EthChainService) dialRPCNode(chainID int) (err error) {
 		RPCEndpoint = "https://api.avax-test.network/ext/bc/C/rpc"
 	case 80001: // Mumbai (Polygon)
 		RPCEndpoint = "https://rpc-mumbai.maticvigil.com"
+	case 4002: // Fantom testnet
+		RPCEndpoint = "https://rpc.testnet.fantom.network"
 	case 1337: //Local Ganache
 		RPCEndpoint = "HTTP://127.0.0.1:8545"
 	default:
--- a/chain/ethChainService.go
+++ b/chain/ethChainService.go
@@ -69,8 +69,7 @@ func (eth *EthChainService) RedeemPhonon(p *model.Phonon, privKey *ecdsa.Private
 	redeemValue := eth.calcRedemptionValue(onChainBalance, suggestedGasPrice)
 	log.Debug("transaction redemption value is: ", redeemValue)
 
-	//If gas would cost more than the value in the phonon, return error
-	if suggestedGasPrice.Cmp(onChainBalance) != -1 {
+	if redeemValue.Cmp(big.NewInt(0)) > 0 {
 		log.Error("phonon not large enough to pay gas for redemption")
 		return "", errors.New("phonon not large enough to pay gas for redemption")
 	}
--- a/chain/ethChainService.go
+++ b/chain/ethChainService.go
@@ -137,6 +137,12 @@ func (eth *EthChainService) dialRPCNode(chainID int) (err error) {
 		RPCEndpoint = "https://eth-rinkeby.gateway.pokt.network/v1/lb/621e9e234e140e003a32b8ba"
 	case 42: //Kovan
 		RPCEndpoint = "https://poa-kovan.gateway.pokt.network/v1/lb/621e9e234e140e003a32b8ba"
+	case 97: // Binance
+		RPCEndpoint = "https://data-seed-prebsc-1-s1.binance.org:8545"
+	case 43113: // Avalanche
+		RPCEndpoint = "https://api.avax-test.network/ext/bc/C/rpc"
+	case 80001: // Polygon
+		RPCEndpoint = "https://rpc-mumbai.maticvigil.com"
 	case 1337: //Local Ganache
 		RPCEndpoint = "HTTP://127.0.0.1:8545"
 	default:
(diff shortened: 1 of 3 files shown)


maevyn11:
Do you happen to have some instructions or handy links for adding all these networks to Metamask? Would be good to provide to the DAO to include with the testnet instructions.

zerodouglas:
Sure! Here are tutorials for MetaMask
- Binance Smart Chain [Binance MM](https://support.binance.us/hc/en-us/articles/4423865255319-Configuring-MetaMask-with-Binance-Smart-Chain)
- Avalanche [Avalanche MM](https://medium.com/seedify/how-to-add-an-avalanche-c-chain-rpc-to-metamask-e8f08979a8a8)
- Polygon [Polygon MM](https://www.coincarp.com/chainlist/mumbai/)

zerodouglas:
Here are faucets as well.

- Binance Smart Chain [Binance Faucet](https://testnet.binance.org/faucet-smart)
- Matic [Mumbai Faucet](https://mumbaifaucet.com/)
- Avalanche [Avalanche Faucet](https://faucet.avax.network/)

chain/ethChainService.go
```diff
@@ -137,6 +137,12 @@ func (eth *EthChainService) dialRPCNode(chainID int) (err error) {
 		RPCEndpoint = "https://eth-rinkeby.gateway.pokt.network/v1/lb/621e9e234e140e003a32b8ba"
 	case 42: //Kovan
 		RPCEndpoint = "https://poa-kovan.gateway.pokt.network/v1/lb/621e9e234e140e003a32b8ba"
+	case 97: // Binance
```
maevyn11, line 140: Would change comment to // Binance Testnet

chain/ethChainService.go
```diff
@@ -137,6 +137,12 @@ func (eth *EthChainService) dialRPCNode(chainID int) (err error) {
...
+		RPCEndpoint = "https://data-seed-prebsc-1-s1.binance.org:8545"
+	case 43113: // Avalanche
```
maevyn11: similarly, could use a comment on which testnet this endpoint points to, seems like there are a bunch of different ones.

chain/ethChainService.go
```diff
@@ -137,6 +137,12 @@ func (eth *EthChainService) dialRPCNode(chainID int) (err error) {
...
+		RPCEndpoint = "https://api.avax-test.network/ext/bc/C/rpc"
+	case 80001: // Polygon
```
maevyn11, line 144: Same as above for this comment

zerodouglas:
I should probably note that I personally was using chain ID 43113 for AVAX on my MetaMask rather than 43114. I see that the tutorial that I provided uses 43114 and so does the phonon-app. I didn't have any issues but I made the change to align with phonon-app.

chain/ethChainService.go
```diff
@@ -196,6 +211,17 @@ func (eth *EthChainService) submitLegacyTransaction(ctx context.Context, nonce u
...
 	}
 
+	balance, err := eth.checkBalance(ctx, redeemAddress.Hex())
+	if err != nil {
+		log.Error("error checking balance: ", err)
+		return signedTx, err
+	}
+
+	if redeemValue.Cmp(eth.calcRedemptionValue(balance, gasPrice)) < 0 {
+		log.Error("balance is insufficient to cover redemption value")
```
maevyn11, line 221: think this should say. Balance is insufficient to cover gas cost of redemption.

chain/ethChainService.go
```diff
@@ -196,6 +211,17 @@ func (eth *EthChainService) submitLegacyTransaction(ctx context.Context, nonce u
...
+		return signedTx, errors.New("balance is insufficient to cover redemption value")
+	}
```
maevyn11, line 223: Actually isn't this whole check redundant with the parent function redeemPhonon starting with the block at line 65? 
```
	//Collect on chain details for redeem
	nonce, onChainBalance, suggestedGasPrice, err := eth.fetchPreTransactionInfo(ctx, common.HexToAddress(p.Address))
	if err != nil {
		return "", err
	}
	redeemValue := eth.calcRedemptionValue(onChainBalance, suggestedGasPrice)
	log.Debug("transaction redemption value is: ", redeemValue)
	//If gas would cost more than the value in the phonon, return error
	if suggestedGasPrice.Cmp(onChainBalance) != -1 {
		log.Error("phonon not large enough to pay gas for redemption")
		return "", errors.New("phonon not large enough to pay gas for redemption")
	}
	tx, err := eth.submitLegacyTransaction(ctx, nonce, big.NewInt(int64(p.ChainID)), common.HexToAddress(redeemAddress), redeemValue, eth.gasLimit, suggestedGasPrice, privKey)
	if err != nil {
		return "", err
	}
```

chain/ethChainService.go
```diff
@@ -196,6 +211,17 @@ func (eth *EthChainService) submitLegacyTransaction(ctx context.Context, nonce u
 		return signedTx, err
 	}
 
+	balance, err := eth.checkBalance(ctx, redeemAddress.Hex())
+	if err != nil {
+		log.Error("error checking balance: ", err)
+		return signedTx, err
+	}
```
maevyn11, line 218: Think this can probably be dropped since this is already handled in fetchPreTransaction issue and not called anywhere if the below check is dropped.

gui/api.go
```diff
@@ -360,6 +360,10 @@ func (apiSession apiSession) redeemPhonons(w http.ResponseWriter, r *http.Reques
 			err:             respErr,
 		})
 	}
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
```
maevyn11, line 366: since the above may loop through several redeems, I think this error check will only catch an error in the last redeem. 

Changing the response status code when there is an error is a good idea, but this should probably still respect the response format and send back the resps slice, which already includes an err field, especially because it could include some successful or partially successful responses. Then again, the error handling for this route is pretty ugly anyway, so maybe this makes no functional difference.  

Maybe 
```
success := true
for _, res := range resps {
  if res.err != nil {
    success = false 
  }
 }
 if !success{
 ...encode with an error status code
 }

**Review by maevyn11 — CHANGES_REQUESTED** [review]
Confirmed working for me with at least AVAX testnet

File: chain/ethChainService.go (hunk 5 above)
Comment (zerodouglas) [reply chain/ethChainService.go:223]: Yes, you're right. I added this because on line 65, it seems to work well for ETH chains only. When I attempt the same test on say Polygon, I receive "rlp: cannot encode negative *big.Int".

File: chain/ethChainService.go (hunk 1 above)
Comment (zerodouglas) [reply chain/ethChainService.go]: Changes made at https://github.com/GridPlus/phonon-client/blob/2f7e312ad4b0cb026611e2f8eefbc960aecf7286/chain/ethChainService.go#L140

File: chain/ethChainService.go (hunk 2 above)
Comment (zerodouglas) [reply chain/ethChainService.go]: I named them based on their RPC node name, would this be better? https://github.com/GridPlus/phonon-client/blob/2f7e312ad4b0cb026611e2f8eefbc960aecf7286/chain/ethChainService.go#L142

File: chain/ethChainService.go (hunk 4 above)
Comment (zerodouglas) [reply chain/ethChainService.go]: I made this change but I'm still curious about my question in response to yours. https://github.com/GridPlus/phonon-client/blob/2f7e312ad4b0cb026611e2f8eefbc960aecf7286/chain/ethChainService.go#L221

File: gui/api.go (hunk 7 above)
Comment (zerodouglas) [reply gui/api.go:373]: That's a great idea and I totally agree. The change is at: https://github.com/GridPlus/phonon-client/blob/2f7e312ad4b0cb026611e2f8eefbc960aecf7286/gui/api.go#L364

File: chain/ethChainService.go (hunk 5 above)
Comment (maevyn11) [reply chain/ethChainService.go:223]: Thought that was just due to the phonon balance having not been high enough?

File: chain/ethChainService.go (hunk 1 above)
Comment (maevyn11) [reply chain/ethChainService.go]: Looks good. And I appreciate the line link

File: chain/ethChainService.go (hunk 2 above)
Comment (maevyn11) [reply chain/ethChainService.go]: perfect

File: gui/api.go [hunk 8]
```diff
@@ -360,8 +360,15 @@ func (apiSession apiSession) redeemPhonons(w http.ResponseWriter, r *http.Reques
...
-	if err != nil {
-		http.Error(w, err.Error(), http.StatusBadRequest)
+
+	success := true
+	for _, res := range resps {
+		if res.err != "" {
+			success = false
+		}
+	}
+	if !success {
```
maevyn11, line 370: This will still only send back the last error, although it now detects errors better. This should ideally encode and send the whole resps as in enc.Encode() below but set the status code to http.StatusInternal error if !success

maevyn11 (changes requested):

gui/api.go
zerodouglas: Changes for this begin at https://github.com/GridPlus/phonon-client/blob/847ba4d15d383cf8a2faac8dc4530601fd428d8a/gui/api.go#L345

zerodouglas:
Fantom Testnet information as follows:

- [MetaMask Tutorial](https://docs.fantom.foundation/tutorials/set-up-metamask)
- [Faucet](https://faucet.fantom.network)

note* Has been added into the phonon-app.

gui/api.go
maevyn11: If there are multiple errors does this append duplicate headers or overwrite them? If the former this should ideally be prevented. 

Also I believe the swagger.yml needs to be updated to reflect this API change. I think the "200" code sections happens to already be unparseable by swagger as well, so that ought to be remedied as well.

zerodouglas: Ah I missed this. I wouldn't think you could or want to override them especially in this range. I'll add back the variable to set the boolean as it actually makes big sense. I'll update the swagger now aswell :)

zerodouglas: Majority is the random formatting that occurred, otherwise, the changes start at https://github.com/GridPlus/phonon-client/blob/5abf9563ae242bc7071e9f7880ad45f4056ee391/gui/swagger.yaml#L467. You should now be able to get a parsed response.

maevyn11: looks good now

maevyn11 (approved):
Looks great. Really fun to see this feature working.

PR #205, opened 2022-08-09
Fix api cached pubkeys

maevyn11:
Prevented the app from unnecessarily waiting for the API to make a bunch of GET_PHONON_PUB_KEY calls when the keys are already cached. 

Also deleted or downgraded a bunch of pointless or obscure logs.


gui/api.go
```diff
@@ -638,10 +649,12 @@ func (apiSession apiSession) listPhonons(w http.ResponseWriter, r *http.Request)
 	}
 
 	for _, p := range phonons {
-		p.PubKey, err = sess.GetPhononPubKey(p.KeyIndex, p.CurveType)
-		if err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
-			return
+		if p.PubKey == nil {
```
npmaile, line 652: good catch. I didn't even think about this or notice it.

npmaile (approved):

chain/ethChainService.go
```diff
@@ -185,6 +190,8 @@ func (eth *EthChainService) calcRedemptionValue(balance *big.Int, gasPrice *big.
 	return valueMinusGas.Sub(balance, estimatedGasCost.Mul(gasPrice, big.NewInt(int64(gasLimit))))
 }
 
+func (eth *EthChainService) IsRedeemable()
```
npmaile, line 193: no function body. oops

npmaile (changes requested):

chain/ethChainService.go
maevyn11: oops. I moved that change to a separate branch, forgot to delete before committing this one. Will remove.

maevyn11: https://github.com/GridPlus/phonon-client/pull/205/commits/580f2a21c8a451a1f576d62237ec6c145639d370

gui/api.go
maevyn11: neither did I until I was sitting waiting for the page to load while 15 GET_PHONON_PUB_KEYS ran :)

npmaile (approved):

PR #206, opened 2022-08-10
Fix destroying unredeemable phonons

maevyn11:
Added a checkRedeemable method to the chain interface so that we can check whether the phonon is going to be able to be redeemed before destroying it to reveal it's private key. 

Copy pasted some code and didn't unit test because I expect most of the EthChainService module will get thrown away as build the new version for the Beta. I did run through successful and unsuccessful redeem checks and confirmed it works as intended. 

Still factored some things into helper methods and added standardized Error variables because I can't help myself.


zerodouglas (approved):
lgtm. tested as well (check)

PR #207, opened 2022-08-11
Standalone jumpbox

npmaile:
pretty simple. Make a make target and go file for building only the jumpbox without needing any non-go nonesense

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .vscode/
+jumpbox
 phonon
 phonon.exe
 localCerts/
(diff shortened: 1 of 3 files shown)


maevyn11:
What's the purpose of this? Seems like this is redundant with the cmd/JumpBox.go and doesn't support the same configuration options?

npmaile:
> What's the purpose of this? Seems like this is redundant with the cmd/JumpBox.go and doesn't support the same configuration options? 

it produces a slimmer binary and doesnt require frontend, ayatana-appindicator, or gtk3

currently necessary for building on amazon linux

maevyn11 (approved):

PR #209, opened 2022-08-15
Winstaller

npmaile:
creates a windows installer


npmaile:
can't go in until the better-systray goes in

maevyn11 (approved):

PR #210, opened 2022-08-16
inital mining operation endpoint

zerodouglas:
Adds support for mining native Phonons in the GUI.


orchestrator/session.go
```diff
@@ -113,7 +144,96 @@ func (s *Session) handleIncomingSessionRequests() {
...
+	return ErrMiningNotActive
+}
+
+func (s *Session) MineNativePhonon(difficulty uint8) error {
+	if !s.verified() {
+		return card.ErrPINNotEntered
+	}
+
+	s.ElementUsageMtex.Lock()
+	defer s.ElementUsageMtex.Unlock()
```
maevyn11, line 176: I think this lock needs to be moved inside of the go func. This defer statement is going to run when the enclosing function (starting on line 170) exits, but the goroutine anonymous function will still be running MineNative in a loop, so I think the lock will released before mining has even really started. 

Ideally this lock would enclose the operation on line 199 s.cs.MineNativePhonon to ensure no other session command can come in and try to communicate with the card in another thread.

orchestrator/session.go
```diff
@@ -113,7 +144,96 @@ func (s *Session) handleIncomingSessionRequests() {
...
+					TimeElapsed: int64(totalTime.Milliseconds()),
+					StartTime:   start,
+					StopTime:    time.Now(),
+				}
+				return
+			default:
+				s.isMiningActive = true
+				keyIndex, hash, err := s.cs.MineNativePhonon(difficulty)
+				elapsed := time.Since(start)
+				totalTime += elapsed
```
maevyn11, line 201: I think this value may be incorrect? Adding up elapsed on each iteration will repeatedly accumulate the time from every previous loop, basically doublecounting each prior iteration on every new iteration. 

I think `elapsed` is the correct total time here and `totalTime` can just be replaced by it in the report output.

orchestrator/session.go
```diff
@@ -113,7 +144,96 @@ func (s *Session) handleIncomingSessionRequests() {
...
+				} else if err != nil {
+					s.miningReportStatus = &miningReportStatus{
+						Attempts:    i,
+						Status:      StatusMiningError,
+						TimeElapsed: int64(totalTime.Milliseconds()),
+						StartTime:   start,
+					}
+					return
+				} else {
+					averageTime := time.Duration(float64(totalTime.Nanoseconds()) / float64(i))
```
maevyn11, line 218: I think average time could be reported in the other two cases as well. It'd just be average time of each iteration so far.

orchestrator/session.go
```diff
@@ -113,7 +144,96 @@ func (s *Session) handleIncomingSessionRequests() {
...
 
-func (s *Session) SetPaired(status bool) {
+func (s *Session) SetPaired(status bool) {}
+
+func (s *Session) GetMiningReport() (*miningReportStatus, error) {
+	if !s.verified() {
+		return nil, card.ErrPINNotEntered
+	}
+
+	return s.miningReportStatus, nil
```
maevyn11, line 154: Since this value is initialized to nil this statement could panic. Needs to check `if s.miningReportStatus != nil` or the value needs to be initialized to some reasonable default. Not sure there is a reasonable default in the case that mining hasn't taken place yet so the nil check is probably cleanest.

maevyn11 (changes requested):
Few small issues to cover, and needs the frontend submodule included before merging, but the bones look good.

gui/api.go
```diff
@@ -249,6 +252,72 @@ func (apiSession apiSession) createPhonon(w http.ResponseWriter, r *http.Request
...
+		return
+	}
+
+	err = sess.MineNativePhonon(req.Difficulty)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
```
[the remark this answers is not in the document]
zerodouglas: I agree. I made that change here https://github.com/GridPlus/phonon-client/blob/a8a58b5610af6fa9e412692e5e275ddadcff4ac2/gui/api.go#L316

orchestrator/session.go
```diff
@@ -30,6 +34,7 @@ type Session struct {
 	remoteMessageChan     chan (model.SessionRequest)
 	remoteMessageKillChan chan interface{}
 	active                bool
+	hash                  []byte
```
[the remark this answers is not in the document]
zerodouglas: Correct. It appears to be unused, I've removed it. https://github.com/GridPlus/phonon-client/blob/a8a58b5610af6fa9e412692e5e275ddadcff4ac2/orchestrator/session.go#L29

orchestrator/session.go
```diff
@@ -38,10 +43,31 @@ type Session struct {
 	logger                *log.Entry
 	chainSrv              chain.ChainService
 	cache                 map[model.PhononKeyIndex]cachedPhonon
+	cancelMiningChan      chan bool
+	isMiningActive        bool
+	miningReportStatus    *miningReportStatus
```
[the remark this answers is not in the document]
zerodouglas: I like this name better; refer to the above comment for location of change.

orchestrator/session.go
```diff
@@ -68,13 +95,17 @@ func NewSession(storage model.PhononCard) (s *Session, err error) {
 		remoteMessageChan:     make(chan model.SessionRequest),
 		remoteMessageKillChan: make(chan interface{}),
 		active:                true,
+		hash:                  nil,
```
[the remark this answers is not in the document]
zerodouglas: Removed as well.

orchestrator/session.go
zerodouglas: Hmm, I missed this one. It does so you've described, I've moved this into the go func and it seems to be resolved, I hope this is what you were referring to. https://github.com/GridPlus/phonon-client/blob/a8a58b5610af6fa9e412692e5e275ddadcff4ac2/orchestrator/session.go#L182

orchestrator/session.go
zerodouglas: I partly understand. It could be the time of day but I went ahead and removed the totalTime and rather instead used elapsed and used the supported methods to make it cleaner. https://github.com/GridPlus/phonon-client/blob/a8a58b5610af6fa9e412692e5e275ddadcff4ac2/orchestrator/session.go#L190

orchestrator/session.go
zerodouglas: Agree, I've made the change to be in the two other cases!

orchestrator/session.go
zerodouglas: The endpoint will now return "no mining status report available" if the struct is empty. https://github.com/GridPlus/phonon-client/blob/a8a58b5610af6fa9e412692e5e275ddadcff4ac2/orchestrator/session.go#L152

orchestrator/session.go
maevyn11: Yea, actually not exactly what I had in mind but I think your implementation is better. I was thinking of locking and unlocking on each iteration, basically like this within the for loop: 
```
s.ElementusageMtex.Lock()
s.cs.MineNativePhonon
s.ElementUsageMtex.Unlock()
```
Putting it at the top of the anon func with defer is nice though, the lock will be held during the entire mining attempt, so there's no chance of other commands being slipped in between unexpectedly. So yea this is better, nice.

orchestrator/session.go
maevyn11: Think this looks correct now

orchestrator/session.go
```diff
@@ -44,7 +46,7 @@ type Session struct {
 	cache                 map[model.PhononKeyIndex]cachedPhonon
 	cancelMiningChan      chan bool
 	isMiningActive        bool
-	miningStatusReport    *miningStatusReport
+	miningStatusReport    map[string][]*miningStatusReport
```
maevyn11, line 49: I think this could just a map to a single MiningStatusReport, instead of a slice. So `map[string]*miningStatusReport` 

This data could grow large pretty quickly when storing a report for each iteration. Unless we want to be able to retrieve reports from prior iterations? Thoughts @italianst4 ?

maevyn11 (commented):
One open ended question to handle and then I think this is good to go.

orchestrator/session.go
npmaile: we should just keep the last status report. No need to store the previous data as we are storing aggregate statistics

orchestrator/session.go
```diff
@@ -75,6 +104,9 @@ func NewSession(storage model.PhononCard) (s *Session, err error) {
 		ElementUsageMtex:      sync.Mutex{},
 		logger:                log.WithField("CardID", "unknown"),
 		chainSrv:              chainSrv,
+		cancelMiningChan:      make(chan bool),
```
npmaile, line 107: chan struct{} is smaller by like a bit

npmaile (changes requested):
I'd drop tiny comments all over the place, but the issue here is that you can't look at past mining reports. The id of the mining reports, and their corresponding reports need to be stored in the gui package. This way, if someone slams the mine button 20 times, they can retrieve the status reports of all twenty mining operations including the eventual hashes from them. assigning an id for each mining attempt(as defined by the user clicking the "mine" button) should happen in the gui package.

orchestrator/session.go
zerodouglas: Results at https://github.com/GridPlus/phonon-client/blob/3ba229ee70798a3e6abf252ec1ab8986d2c6f495/orchestrator/session.go#L49

orchestrator/session.go
```diff
@@ -113,7 +143,146 @@ func (s *Session) handleIncomingSessionRequests() {
...
+		defer s.ElementUsageMtex.Unlock()
+
+		start := time.Now()
+		for {
+			select {
+			case <-cancel:
+				s.isMiningActive = false
+				elapsed := time.Since(start)
+				log.Debug("mining successfully cancelled")
+				s.miningStatusReport[id] = &miningStatusReport{
```
npmaile, line 221: I'd probably create one mining Status Report, and then overwrite the fields as the process is going along

npmaile: rather than create one fresh each time one is being used
though you then have to have the map point to a concrete miningStatusReport object instead of a pointer because you could get concurrent read while another thread is writing, which would panic.

orchestrator/session.go
douglance: the frontend doesn't want to be responsible for uuids (if thats what you mean by id here)

gui/api.go
```diff
@@ -89,7 +89,8 @@ func Server(port string, certFile string, keyFile string, mock bool) {
 	r.HandleFunc("/cards/{sessionID}/phonon/{PhononIndex}/export", session.exportPhonon)
 	r.HandleFunc("/cards/{sessionID}/phonon/mineNative", session.mineNativePhonons)
 	r.HandleFunc("/cards/{sessionID}/phonon/mineNative/cancel", session.cancelMineRequest)
-	r.HandleFunc("/cards/{sessionID}/phonon/mineNative/status", session.statusMineRequest)
+	r.HandleFunc("/cards/{sessionID}/phonon/mineNative/status", session.listMiningReportStatus)
+	r.HandleFunc("/cards/{sessionID}/phonon/mineNative/status/{miningSessionID}", session.miningReportStatus)
```
maevyn11, line 93: miningSessionID is a little confusing with sessionID already in existence. Maybe we can use a better distinguished variable name? My ideas below:

miningAttemptID
miningRunID

gui/api.go
```diff
@@ -851,3 +883,12 @@ func (apiSession apiSession) sessionFromMuxVars(p map[string]string) (*orchestra
 	}
 	return targetSession, nil
 }
+
+func sessionFromMiningVars(p map[string]string) (string, error) {
```
maevyn11, line 887: Think this should be renamed "miningSessionIDFromVars" as there are no "mining vars", just path variables we're parsing here. 

Also, is this the cleanest way to parse parameters our of URLs? I genuinely don't know, it just feels like this pattern is a little clunky to add a helper function with redundant error checking every time we parse a variable out of a URL.

gui/api.go
```diff
@@ -851,3 +883,12 @@ func (apiSession apiSession) sessionFromMuxVars(p map[string]string) (*orchestra
...
+	miningSessionName, ok := p["miningSessionID"]
+	if !ok {
+		fmt.Println("unable to find mining session")
```
maevyn11, line 890: This should be a log.Error instead of a Println. Generally shouldn't have println anywhere in the codebase outside of terminal UI's

maevyn11: oh yea and my apdu.log hack... 😬

gui/swagger.yaml
```diff
@@ -417,6 +422,31 @@ paths:
...
         description: sessionID of connected card
+  /cards/{sessionID}/phonon/mineNative/status/{miningSessionID}:
+    get:
+      tags:
+        - "sessions"
+      responses:
+        "500":
+          description: "no mining status report available"
+        "404":
+          description: "no session with id"
```
maevyn11, line 433: What situation would lead to each of these errors? Feels like we'd always hit the 500 if this request was made to a card that hasn't mined yet. Maybe this should be one error code?

orchestrator/session.go
maevyn11: good point. Apparently the semantics around concurrent read/write to maps has changed since go 1.6. TLDR it used to be safe to have just one writer and as many concurrent readers as you wanted, but now it's not. https://stackoverflow.com/questions/36167200/how-safe-are-golang-maps-for-concurrent-read-write-operations 

Seems like the proper way to do this now is to use a thread safe map from the sync package. See here: https://golangprojectstructure.com/maps-thread-safe-goroutine-concurrency/

The way this currently works I think we'll get rare inconsistent results, like undefined behavior when reading the status as another attempt is being written, and possibly panics, so think we need to resolve this.

orchestrator/session.go
```diff
@@ -251,15 +260,29 @@ func (s *Session) MineNativePhonon(difficulty uint8) error {
...
 						Hash:        hex.EncodeToString(hash),
-					})
+					}
+					phonons, err := s.cs.ListPhonons(0, 0, 0, false)
+					if err != nil {
+						fmt.Println("error listing phonons: ", err)
+						return
+					}
+					for _, p := range phonons {
+						if p.KeyIndex == keyIndex {
```
maevyn11, line 270: This change to check the keyIndex looks good, but I think you still have to call GetPhononPubkey and assign the pubkey or the frontend won't be able to see the pubkey/hash. 

```
pubkey, err := s.cs.GetPhononPubKey()
p.PubKey = pubkey
```

maevyn11: actually, scratch that, this information is already in the function scope from the MineNativePhonon call. So we can skip the GetPhononPubKey call and just do this instead: 
```
p.PubKey = hash
```

orchestrator/session.go
```diff
@@ -176,73 +186,72 @@ func (s *Session) CancelMiningRequest() error {
...
+	cancel := make(chan struct{})
 	s.cancelMiningChan = cancel
 
+	id, err := generateId()
+	if err != nil {
+		return "", err
+	}
+
 	i := 0
 	go func() {
```
maevyn11, line 210: Think it would be good to turn this anonymous function into a named one, like initMiningAttempt and call that here via go initMiningAttempt instead. Would help separate the logic for generating the id, starting the attempt, and returning it from the actual logic of doing the miningAttempt and tracking the status.

maevyn11 (changes requested):
Just a few comments on things that could be simplified/clarified and a possible race condition to work out. Overall I think the layout for handling miningAttemptIDs between the Session and the gui API looks really good.

maevyn11:
Also, this branch needs to check in the frontend submodule before merging.

gui/api.go
zerodouglas: Sure! I like miningAttemptID

gui/api.go
zerodouglas: Noted, I'll also update sessionFromMuxVars following this as well

gui/api.go
zerodouglas: Mux recommends to retrieve them by calling `mux.Vars()`. I could instead of using a helper function, do a 1 liner in the function that needs to retrieve the mining attempt id with `attemptId := vars["miningAttemptID"]`. Regardless if there is no id in the path variable, the server responds with "404 not found" and with no trailing slash, responds with the list of statuses.

orchestrator/terminal.go
```diff
@@ -5,18 +5,13 @@ import (
...
 	"github.com/GridPlus/phonon-client/card"
-	"github.com/GridPlus/phonon-client/model"
 	"github.com/GridPlus/phonon-client/usb"
 )
 
 type PhononTerminal struct {
 	sessions []*Session
 }
 
-type remoteSession struct {
```
npmaile, line 16: was this unused?

zerodouglas: Yes, on the master branch, my linter is showing that this is unused.

gui/api.go
maevyn11: Cool, thanks for the explanation, was worried that this statement could panic
```
vars := mux.Vars(r)
attemptId := vars["miningAttemptID"]
```

But seems like mux.Vars(r) will always return a miningAttemptID as long as it was included in the URL, so this panic is impossible. Kind of awkward to rely on this implicit guarantee, but I guess as long as we remember that's how mux.Vars works its cleaner than all the redundant error handling.

gui/swagger.yaml
```diff
@@ -1,655 +1,627 @@
-swagger: "2.0"
+openapi: 3.0.0
```
maevyn11, line 1: What does this change?

orchestrator/session.go
```diff
@@ -193,94 +209,102 @@ func (s *Session) CancelMiningRequest() error {
...
+	if !s.verified() {
+		return "", card.ErrPINNotEntered
+	}
+
+	id, err := generateId()
+	if err != nil {
+		return "", err
+	}
+
+	go s.initMiningAttempt(id, difficulty)
```
maevyn11, line 307: Nice, this one looks very tidy now

orchestrator/session.go
```diff
@@ -193,94 +209,102 @@ func (s *Session) CancelMiningRequest() error {
...
+							fmt.Println("error getting public key: ", err)
+							return
+						}
+
+						p.PubKey = pubkey
+
+						s.cache[keyIndex] = cachedPhonon{
+							pubkeyCached: true,
+							infoCached:   true,
+							p:            p,
```
maevyn11, line 286: Pubkey setting looks good now

orchestrator/session.go
```diff
@@ -131,6 +136,21 @@ func NewSession(storage model.PhononCard) (s *Session, err error) {
...
+
+func (m mutexedMiningReport) getMiningStatus(attemptId string) (miningStatusReport, error) {
+	m.mtex.Lock()
+	defer m.mtex.Unlock()
+	if _, ok := m.m[attemptId]; ok {
+		return m.m[attemptId], nil
+	}
+	return miningStatusReport{}, ErrMiningReportNotAvailable
+}
+
```
maevyn11, line 153: Nice, this whole structure is a nice clean implementation for the needed locking.

maevyn11 (approved):
Change looks really good. Just wondering why all the format updates to the swagger.yml?

gui/swagger.yaml
zerodouglas: There's many properties that were changed with 3.0 than in 2.0. I figured as we're updating documentation, Go version and other resources that it may be best to update the Swagger as well. For instance, produces and consumes was replaced with the new components structure. Hosts, Schemes, and BasePath was replaced with servers.

maevyn11: Gotcha, yea updating seems good then. Doesn't seem like it changes too much but usually pays to stay up to date.

orchestrator/session.go
```diff
@@ -113,7 +163,150 @@ func (s *Session) handleIncomingSessionRequests() {
...
+	i := 0
+	s.ElementUsageMtex.Lock()
+	defer s.ElementUsageMtex.Unlock()
+
+	report := miningStatusReport{}
+
+	start := time.Now()
+	for {
+		select {
+		case <-cancel:
```
npmaile, line 225: this can be cleaned up further yet. I'm not going to make you do it now, but consider it if you have nothing going on sometime

npmaile (approved):

PR #219, opened 2022-08-25
fix card reload fix

npmaile:
fixes the card reload issue. Assets and static content are served the same way, but any path other than the api and swagger paths just serve index.html. This is what I believe is supposed to be the way it works.

### DIFF
--- a/gui/api.go
+++ b/gui/api.go
@@ -23,11 +23,23 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-//go:embed frontend/build/*
-var frontend embed.FS
+//go:embed frontend/build/assets/*
+var frontendAssets embed.FS
+
+//go:embed frontend/build/static/*
+var frontendStatic embed.FS
 
 //go:embed swagger.yaml
 var swaggeryaml []byte
+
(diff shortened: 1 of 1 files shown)


maevyn11:
Nice, looks like this fixes getting the white screen 404 pages when hitting back, forward, or refresh. 

Refresh seems to now work as expected, triggering a reload of the current page, issuing card commands as though it was the first time called, so think that's good. 

Hitting back and forward appear to change the URL but don't change what's displayed on the page, is that the intended behavior?

npmaile:
> Nice, looks like this fixes getting the white screen 404 pages when hitting back, forward, or refresh.
> 
> Refresh seems to now work as expected, triggering a reload of the current page, issuing card commands as though it was the first time called, so think that's good.
> 
> Hitting back and forward appear to change the URL but don't change what's displayed on the page, is that the intended behavior?

not sure. I think that behavior is a purely frontend thing in a spa like this one

maevyn11 (approved):
Looks good, nice straightforward improvement. We can figure out the perfect behavior for the back and forward buttons later.

PR #220, opened 2022-08-26
add check in gui api for telemetry key

npmaile:
this is sort of implemented in a hacky way as the configs are being parsed and updated each time the card session is created, and not in the session where it probably should be. Given that that will require considerable effort and the telemetry is only for testnet, this implementation is fine.

### DIFF
--- a/gui/api.go
+++ b/gui/api.go
@@ -15,6 +15,7 @@ import (
 	"strconv"
 	"time"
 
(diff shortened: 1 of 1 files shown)


maevyn11:
Hack is ugly and good, needs a frontend to go along. Maybe a good community issue to throw out to a frontend dev?

maevyn11 (approved):

PR #230, opened 2022-09-12
actually mine native phonons on mock

npmaile:


zerodouglas (approved):
Very simple indeed, makes sense. LGTM

PR #231, opened 2022-09-12
adds testnet supported chains to be returned at endpoint

zerodouglas:
This PR references the list of chains in the GUI's frontend and provides support to return the specific list of testnet chains from an endpoint called `/listchains`.

### DIFF
--- a/chain/ethChainService.go
+++ b/chain/ethChainService.go
@@ -39,18 +39,18 @@ type supportedChains struct {
 	CurrencyType uint
 }
 
+var chains = []supportedChains{
+	{ChainID: 3, Ticker: "ETH", Name: "Ropsten", CurrencyType: 2},
+	{ChainID: 42, Ticker: "ETH", Name: "Kovan", CurrencyType: 2},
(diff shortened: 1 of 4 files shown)


chain/ethChainService.go
```diff
@@ -32,6 +32,28 @@ type EthChainService struct {
...
+	ChainID      uint
+	Ticker       string
+	Name         string
+	CurrencyType uint
+}
+
+func GetSupportedChains() []supportedChains {
+	var chains []supportedChains
+
+	chains = append(chains, supportedChains{ChainID: 3, Ticker: "ETH", Name: "Ropsten", CurrencyType: 2})
```
npmaile, line 45: Do this as at the top of the file at global(to this package) scope instead of building the list each time you run the function

npmaile (changes requested):

chain/ethChainService.go
zerodouglas: Changes at https://github.com/GridPlus/phonon-client/blob/60957baabed5c1c955d3238aa2f07da821d5c997/chain/ethChainService.go#L42

npmaile (approved):

PR #292, opened 2022-10-03
adding useful error message when installCardCert fails due to attempt…

maevyn11:
…ing to install a cert that doesn't match the CA. The small fix we found while updating regression tests last week.


zerodouglas (approved):
Lgtm. Useful to have!

npmaile (approved):

PR #300, opened 2022-10-09
Issue-239

### LINKED ISSUE #239: Root Cause and resolve ChainID Missing Issue after local app restart
Several testnet users have reported that the chainID of their phonon assets gets lost and appears as empty or undefined after restarting the app. This appears to be a distinct issue from the malformed chainIDs which appear after sending a phonon. 

Reproduction steps: 
1. Create several phonon on various ETH and Testnet chains. 
2. Restart the entire phonon-app. 
3. Observe in the UI that chainIDs may appear incorrect

ahinchliff:
Closes #239 

I've identified why chainId is being misreported for some chains after client restart.

The correct chainId is added to the cache but was not being stored correctly on the card. On client reset the cache is cleared breaking the UI. 

I have supplied a fix but not sure if this is the best solution. Should at least give someone with more knowledge a starting point.

### DIFF
--- a/card/phonon.go
+++ b/card/phonon.go
@@ -10,10 +10,10 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-//TLV Encodes the phonon standard schema used for setting it's descriptor. Must be extended with additional fields
-//to suit the various commands that deal with phonons.
-//Excludes fields KeyIndex, PubKey, and CurveType which are already known by the card at the time of creation
-//Includes fields SchemaVersion, ExtendedSchemaVersion, CurrencyType, Denomination, and ExtendedTLVs
+// TLV Encodes the phonon standard schema used for setting it's descriptor. Must be extended with additional fields
+// to suit the various commands that deal with phonons.
+// Excludes fields KeyIndex, PubKey, and CurveType which are already known by the card at the time of creation
+// Includes fields SchemaVersion, ExtendedSchemaVersion, CurrencyType, Denomination, and ExtendedTLVs
 func TLVEncodePhononDescriptor(p *model.Phonon) ([]byte, error) {
 	//KeyIndex omitted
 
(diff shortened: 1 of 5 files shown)


card/phonon.go
```diff
@@ -50,10 +50,13 @@ func TLVEncodePhononDescriptor(p *model.Phonon) ([]byte, error) {
 		return nil, err
 	}
 	//adding chainID support through extended schema to avoid card code change
-	chainIDTLV, err := tlv.NewTLV(TagChainID, []byte{byte(p.ChainID)})
+	chainIdBytes := make([]byte, 4)
+	binary.BigEndian.PutUint32(chainIdBytes, p.ChainID)
+	chainIDTLV, err := tlv.NewTLV(TagChainID, chainIdBytes)
```
maevyn11, line 55: Nice! So this is the real fix right here right? I can see now that []byte{byte(p.ChainID)}} would not work reliably, maybe it did for values less than one byte or something. 

Is it safe to make ChainID a uint32 though or will we run into problems representing certain chainIDs?

maevyn11 (changes requested):
Fix makes sense. Just one question about the datatype.

card/phonon.go
ahinchliff: My thinking was that a max value of 4.2B would be sufficient to cover future chain ids but maybe that is too much of an assumption. Let me know if you want me to change this back.

maevyn11: ah no, I think I got my byte math way wrong at first look. I think 4.2B max value is more than sufficient.

maevyn11 (approved):
Good catch and thanks for the fix! Seems the caching layer hid this issue from view in the test phase. Fix looks good to me. 

@zerodouglas any chance this issue played a part in the chainID issues when sending to cards. It looks like to me that with the current implementation any chainID > 1 byte in length would've been mangled on input and output to the card.

ahinchliff:
No worries. Just a reminder I dont have merge permissions so someone else will need to do the honours.

PR #306, opened 2022-10-17
Move config

npmaile:
oops. I noticed there was a little bit of formatting that changed during my change. I figured one of two things is true. 

1. my formatter is messed up and running go fmt will fix it
2. somewhere in the code is messed up and go fmt will fix it

the second was correct.

### DIFF
--- a/repl/repl.go
+++ b/repl/repl.go
@@ -23,7 +23,13 @@ func Start() {
 	shell = ishell.New()
 	conf, err := config.LoadConfig()
 	if err != nil {
-		log.Fatal("Unable to load configuration")
+		fmt.Printf("configuration loading failed with error %s\nWould you like to continue?\n(y/n): ", err.Error())
+		var res string
+		fmt.Scanln(&res)
+		if strings.ToLower(res)[0:0] != "y" {
+			log.Fatal("exiting")
+		}
+
 	}
 	t = orchestrator.NewPhononTerminal(conf)
 	// get initial state of orchestrator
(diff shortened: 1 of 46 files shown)


repl/repl.go
```diff
@@ -19,7 +21,11 @@ const standardPrompt string = "Phonon>"
 
 func Start() {
 	shell = ishell.New()
-	t = orchestrator.NewPhononTerminal()
+	conf, err := config.LoadConfig()
```
maevyn11, line 24: Shouldn't this use the new config.MustLoadConfig()?

maevyn11 (changes requested):
Looks like this combines the go fmt changes and the config update? The conf reshuffle is long overdue and looks good. I just found one place that wasn't consistent with the new style that should probably be updated.

repl/repl.go
npmaile: no, I want to check config errors, and report them back instead of killing the program immediately. It could go either way, but the MustLoadConfig was just to kill early on usage of the command line scripts, and not really meant for a frontend which will likely have it's own error handling/mitigation

maevyn11: Won't log.Fatal() kill the program anyway though? If this is the intent then that should probably be log.Error() instead.

npmaile: yes, but it's the frontend doing it and not the config parser.

maevyn11: Why does it matter if the frontend kills the program or the config? Either way a bad config leads to a dead program and the exact same log message is reported? May as well have a consistent pattern for fetching the config.

maevyn11 (approved):

PR #308, opened 2022-10-21
Fix remote

npmaile:
fixes a couple of issues with the remote server
- disconnect not being communicated to the clients properly
- disconnect method on client implemented
- create channel only when it's needed for connectCard2Card message

### DIFF
--- a/orchestrator/session.go
+++ b/orchestrator/session.go
@@ -973,3 +973,12 @@ func (s *Session) addInfoToCache(p *model.Phonon) {
 	}
 
 }
+
+func (s *Session) DisconnectRemote() error {
+	if s.RemoteCard == nil {
+		return nil
+	}
+	err := s.RemoteCard.Disconnect()
+	s.RemoteCard = nil
+	return err
+}
--- a/orchestrator/session.go
+++ b/orchestrator/session.go
@@ -626,7 +626,7 @@ func (s *Session) SendPhonons(keyIndices []model.PhononKeyIndex) error {
 }
 
 func (s *Session) ReceivePhonons(phononTransferPacket []byte) error {
-	if !s.verified() && s.RemoteCard != nil {
+	if !s.verified() && s.RemoteCard == nil {
 		return ErrCardNotPairedToCard
 	}
 	s.ElementUsageMtex.Lock()
@@ -707,13 +707,13 @@ func (s *Session) ConnectToCounterparty(cardID string) error {
 		//we shouldn't get this far and still receive this error
 		return err
 	}
-	err = s.PairWithRemoteCard(s.RemoteCard)
+	err = s.PairWithRemoteCard()
 	return err
 
 }
 
-func (s *Session) PairWithRemoteCard(remoteCard model.CounterpartyPhononCard) error {
-	remoteCert, err := remoteCard.GetCertificate()
+func (s *Session) PairWithRemoteCard() error {
+	remoteCert, err := s.RemoteCard.GetCertificate()
 	if err != nil {
 		return err
 	}
@@ -722,7 +722,7 @@ func (s *Session) PairWithRemoteCard(remoteCard model.CounterpartyPhononCard) er
 		return err
 	}
 	log.Debug("sending card pair request")
-	cardPairData, err := remoteCard.CardPair(initPairingData)
+	cardPairData, err := s.RemoteCard.CardPair(initPairingData)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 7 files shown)


orchestrator/session.go
```diff
@@ -598,7 +598,10 @@ func (s *Session) FinalizeCardPair(cardPair2Data []byte) error {
 
 func (s *Session) SendPhonons(keyIndices []model.PhononKeyIndex) error {
 	log.Debug("Sending phonons")
-	if !s.verified() && s.RemoteCard != nil {
+	if !s.verified() {
+		return card.ErrPINNotEntered
+	}
+	if s.RemoteCard == nil {
```
maevyn11, line 604: good catch

orchestrator/session.go
```diff
@@ -707,13 +716,13 @@ func (s *Session) ConnectToCounterparty(cardID string) error {
 		//we shouldn't get this far and still receive this error
 		return err
 	}
-	err = s.PairWithRemoteCard(s.RemoteCard)
+	err = s.PairWithRemoteCard()
```
maevyn11, line 719: Change makes sense since we already have s.RemoteCard being called above anyway. Is it safe to call the below changed functions though if RemoteCard hasn't been set by calling this method yet? Mostly worried about potential panics/surprising behavior.

remote/v1/client/client.go
```diff
@@ -394,7 +394,7 @@ func (c *RemoteConnection) CardPair2(cardPairData []byte) (cardPairData2 []byte,
 
 func (c *RemoteConnection) FinalizeCardPair(cardPair2Data []byte) error {
 	c.sendMessage(v1.RequestFinalizeCardPair, cardPair2Data)
-	if !(c.pairingStatus == model.StatusPaired) {
+	if c.pairingStatus != model.StatusPaired {
```
maevyn11, line 397: ha, nice change.

remote/v1/client/client.go
```diff
@@ -431,8 +431,15 @@ func (c *RemoteConnection) GetCertificate() (*cert.CardCertificate, error) {
 
 func (c *RemoteConnection) ConnectToCard(cardID string) error {
 	c.logger.Info("sending requestConnectCard2Card message")
+	if c.pairingStatus != model.StatusConnectedToBridge {
+		return errors.New("not connected to phonon bridge")
+	}
```
maevyn11, line 436: What if you call this while the card is paired to another card already. Wouldn't this trigger because the status is in a more "advanced" state than StatusConnectedToBridge?

remote/v1/server/server.go
```diff
@@ -122,29 +133,29 @@ func (c *clientSession) process(msg v1.Message) error {
 		c.endSession(msg)
 	case v1.RequestNoOp:
 		c.noop(msg)
-	case v1.RequestIdentify, v1.ResponseIdentify, v1.RequestCardPair1, v1.ResponseCardPair1, v1.RequestCardPair2, v1.ResponseCardPair2, v1.RequestFinalizeCardPair, v1.ResponseFinalizeCardPair, v1.RequestReceivePhonon, v1.MessagePhononAck, v1.RequestVerifyPaired, v1.ResponseVerifyPaired:
-		c.passthrough(msg)
 	case v1.RequestCertificate:
 		c.provideCertificate()
+	default:
+		c.passthrough(msg)
```
maevyn11, line 139: This is a noticeable inclusion. How was passthrough working before? It seems like it was totally missing in the old version.

remote/v1/client/client.go
npmaile: you're welcome

remote/v1/server/server.go
npmaile: I had a long-ass case that had all of the passthrough methods in it. it was like 200 chars long. This is much easier and allows for the addition of more custom requests to passthrough without needing the server to even have knowledge of them.

remote/v1/client/client.go
npmaile: good catch

orchestrator/session.go
npmaile: This would have to be checked in the frontend calling the methods at the moment

remote/v1/client/client.go
npmaile: fix'd

orchestrator/session.go
maevyn11: These can't be called directly from the frontend though right? Has to be passed through session? Maybe they should be made non public or they should have guards in front to check s.RemoteCard != nil?

remote/v1/client/client.go
maevyn11: nice

remote/v1/server/server.go
maevyn11: ah I see it now. Yes this is much cleaner. Very nice.

maevyn11 (approved):
Nice man, looks good to me.

PR #310, opened 2022-10-24
fix the rest of the channels

npmaile:


npmaile:
Added update to frontend module that includes workaround for disconnect-after-send that should fix the issue.

remote/v1/client/client.go
```diff
@@ -226,21 +220,29 @@ func (c *RemoteConnection) process(msg v1.Message) {
 	case v1.MessageError:
 		c.logger.Error(string(msg.Payload))
 	case v1.MessageIdentifiedWithServer:
-		c.identifiedWithServerChan <- true
+		if c.identifiedWithServerChan != nil {
+			c.identifiedWithServerChan <- true
+		}
```
maevyn11, line 225: What happens if the channel is nil?

maevyn11 (approved):
One general question to consider but satisfied it can go in as is.

Issue #305, opened 2022-10-16
Client crashes when listing a mumbai phonon created before 'chainID fix'


ahinchliff:
### Problem

If a mumbai phonon was created before the merge of #300 it's chainID wasnt stored correctly. #300 fixed this but isnt backwards compatible.  Listing one of these faulty chainID phonons causes the client to crash with `panic: runtime error: index out of range [3] with length 1`. 

### Expected Behaviour

Not sure if it is worth adding a fix for this

### Steps to reproduce

1. Revert the merge of #300
2. Start client, unlock & `create`
4. `setDescriptor 10 2 80001 1`
5. Quit client, checkout master, start client, unlock && `list`.

npmaile:
let's skip it.

npmaile:
I take it back.

npmaile:
fixed in pr above

PR #312, opened 2022-10-26
should fix reading shorter than 4 byte chain ids

npmaile:


npmaile:
https://github.com/GridPlus/phonon-client/issues/305

card/phonon.go
```diff
@@ -163,7 +163,20 @@ func TLVDecodePublicPhononFields(phononTLV tlv.TLVCollection) (*model.Phonon, er
...
-			phonon.ChainID = binary.BigEndian.Uint32(entry.Value)
+			v := make([]byte, 4)
+			// pad the beginning with zeros if it's too short
+			if len(entry.Value) < 4 {
+				i := len(entry.Value) - 1
+				j := 0
+				for i >= 0 {
+					v[len(v)-1-j] = entry.Value[i]
+					j++
+					i--
```
maevyn11, line 174: I think this may be broken for length values other than 1, so 2 and 3. 

At 3, i = 2

v[2] = entry.Value[2]
next iteration
v[1] = entry.Value[1]
next iteration
v[0] = entry.Value[0]

Hence the 0 byte will be padded at the end, which will change the value of the decimal, since the most significant bytes get shifted left while padding right instead of getting padded on the left.

maevyn11 (changes requested):

card/phonon.go
maevyn11: But it breaks with 3 and 2?

npmaile: i = 3 -1 = 2
j = 0
v[3] = entry.Value[2]
i = 1
j = 1
v[2] = entry.Value[1]
i = 0
j = 2
v[1] = entry.Value[0]
done
v = [0,entry.value[0],entry.value[1],entry.value[2]

i = 2 -1 = 1
j = 0
v[3] = entry.Value[1]
i = 0
j = 1
v[2] = entry.Value[0]
done
v = [0,0,entry.value[0],entry.value[1]


i = 1-1 = 0
j = 0
v[2] = entry.Value[0]
done
v = [0,0,0,entry.value[0]

npmaile: now that I'm looking at it, the way it should be written is
v := make([]byte,4 -len(entry.Value))
v = append(v,entry.Value...)

card/phonon.go
```diff
@@ -163,7 +163,10 @@ func TLVDecodePublicPhononFields(phononTLV tlv.TLVCollection) (*model.Phonon, er
 	//Collecting ChainID from extended tags pending a more elegant way to do this
 	for _, entry := range phonon.ExtendedTLV {
 		if entry.Tag == TagChainID {
-			phonon.ChainID = binary.BigEndian.Uint32(entry.Value)
+			v := make([]byte, 4-len(entry.Value))
+			v = append(v, entry.Value...)
+			// pad the beginning with zeros if it's too short
+			phonon.ChainID = binary.BigEndian.Uint32(v)
```
maevyn11, line 169: ha, just read your previous comment. There was an off by one error in my brain. But yea this is more readable, nice.

maevyn11 (approved):

PR #313, opened 2022-10-27
Fixes missing arguments in tests

zerodouglas:
As changes were made to the command set, requiring the config. This PR uses the default config provided in the package `config`, for all the test cases.

### DIFF
--- a/card/phononCommandSet_test.go
+++ b/card/phononCommandSet_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/GridPlus/keycard-go/io"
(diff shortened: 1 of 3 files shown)


maevyn11 (approved):
Good change, assuming this gets everything running. 

FYI I'm not sure how many of these tests are still useful and which have gone stale. Planning to take a look soon to prune any deprecated tests out of the repo, make sure these are all unit tests, and delete or move any integration tests into the card repo.

zerodouglas:
I ran a few tests and they did pass. It results to the default config, so you'd have to change the `Certificate` value, if you're using dev certificate.

PR #314, opened 2022-10-28
Adds mine native in repl

zerodouglas:
This PR adds the ability to add the mine native command in the repl. It uses its own specific function as the repl does not utilize channels and or the status report. It simply uses a for loop with messages printing to the console. The repl specifically requires an argument that is an integer, but I can also add a limit to prevent users from "bricking" their card.

### DIFF
--- a/orchestrator/session.go
+++ b/orchestrator/session.go
@@ -309,6 +309,50 @@ func (s *Session) MineNativePhonon(difficulty uint8) (string, error) {
 	return id, nil
 }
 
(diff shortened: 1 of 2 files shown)


npmaile (commented):
figure out if there's a way to make it so that a <C-c> during the mining process can stop it and kick us back to the input prompt

zerodouglas:
Sure, can do. I'll get back to you on that.

PR #315, opened 2022-11-01
Initial core components

italianst4:
Initial core components to new Beta UI


maevyn11 (approved):
Build works, page comes up, happy to merge this into beta as is. We can get into the weeds on it later on in testing.

PR #327, opened 2022-11-09
Dv test cleanup

maevyn11:
Big test housekeeping pass. All tests are now valuable and passing after these changes. Should be a good base to start adding more unit tests where appropriate. 

Removed card integration tests (`phononCommandSet_test.go`) and checked that they've all been superseded by newer tests within the regression suite housed in `phonon-card` directly. 

Toned down some overly chatty log output in other tests, cleaned out dead comments, and removed old tests that were not functioning well or had become irrelevant. In other cases made some corrections to test to reflect API updates that had been missed.


maevyn11:
good idea, done.

PR #344, opened 2022-12-13
Validator initial

zerodouglas:
This is an opinionated, cut-down, very simple validator interface with Eth.


chain/ethChainService.go
```diff
@@ -21,104 +17,62 @@ import (
...
 
-	//Collect on chain details for redeem
-	nonce, onChainBalance, suggestedGasPrice, err := eth.fetchPreTransactionInfo(ctx, common.HexToAddress(p.Address))
-	if err != nil {
-		return "", err
-	}
-	redeemable, redeemValue := eth.checkRedeemValue(onChainBalance, suggestedGasPrice)
-	if !redeemable {
-		return "", ErrGasCostExceedsRedeemValue
+	if phonon.Address != address {
```
maevyn11, line 52: What is this comparing exactly? That the address derived from the passed in phonon's pubkey is equal to the address previously stored in phonon.Address? Are we sure that value is always guaranteed to be there?

validator/validator.go
```diff
@@ -2,12 +2,57 @@ package validator
...
+	result := []*ProposalValidationResult{}
+
+	if len(proposal) == 0 {
+		return nil, errors.New("there were no phonon(s) to validate")
+	}
+
+	for _, phonon := range proposal {
+		switch phonon.ChainID {
+		case 1:
+			ethChainService := chain.NewEthChainService()
```
maevyn11, line 41: Think this will reinstantiate a new EthChainService object every time you call an ETH phonon to validate, including new RPC node connection?

 I think it'd be better to call chain.NewEthChainService above in NewValidator and save it as struct member so that we don't have to reestablish the RPC connection for every phonon.

chain/ethChainService.go
```diff
@@ -21,104 +17,62 @@ import (
...
-
-	return ethchainSrv, nil
+func NewEthChainService() *EthChainService {
+	return &EthChainService{}
 }
 
-// Derives an ETH address from a phonon's ECDSA Public Key
-func (eth *EthChainService) DeriveAddress(p *model.Phonon) (address string, err error) {
-	eccPubKey, err := model.PhononPubKeyToECDSA(p.PubKey)
+func (eth *EthChainService) Validate(phonon *model.Phonon) (valid bool, err error) {
```
maevyn11, line 34: Shouldn't this Validate method implement the the Validation interface you defined? 
`Validate(phonons *[]model.phonon) ([]*ProposalValidationResult, error)`

validator/validator.go
```diff
@@ -2,12 +2,57 @@ package validator
...
+			valid, err := ethChainService.Validate(phonon)
```
maevyn11, line 42: As I mentioned in this method's definition. I think this ethChainService.Validate() should follow the AssetValidator interface and then the whole phonon list can be passed in without changes or iteration.

validator/validator.go
```diff
@@ -2,12 +2,57 @@ package validator
...
+
+func (v *validator) Validate(proposal []*model.Phonon) ([]*ProposalValidationResult, error) {
+	result := []*ProposalValidationResult{}
+
+	if len(proposal) == 0 {
+		return nil, errors.New("there were no phonon(s) to validate")
+	}
+
+	for _, phonon := range proposal {
+		switch phonon.ChainID {
```
maevyn11, line 39: Think this switch needs to handle CurrencyType before ChainID, no guarantee that the ChainID is on ETH, it could refer to another blockchain entirely.

maevyn11: In fact ChainID should probably be handled internally by each implementation, because the validation logic should be basically the same for every ChainID within a CurrencyType.

validator/validator.go
```diff
@@ -2,12 +2,57 @@ package validator
...
+		case 1:
+			ethChainService := chain.NewEthChainService()
+			valid, err := ethChainService.Validate(phonon)
+			if err != nil {
+				return nil, err
+			}
+
+			result = append(result, &ProposalValidationResult{
```
maevyn11, line 47: along with above, think this result should be returned directly from the Validator implementation so it doesn't have to be compiled here for every different CurrencyType.

validator/validator.go
```diff
@@ -2,12 +2,57 @@ package validator
...
+// ProposalValidationResult is the data that is returned from the validator response.
+type ProposalValidationResult struct {
+	phonon *model.Phonon
+	valid  bool
+	err    error
+}
 
-//Validates that a phonon's presented public key represents an actual crypto asset
+// Validator is the interface to validating phonon(s) assets on a given chain.
 type Validator interface {
```
maevyn11, line 18: Think AssetValidator would be a better name for the interface, Validator does not specify what is being validated and could refer to anything.

maevyn11 (changes requested):
Left some comments on the validator and ethChainService changes. Tough to review with all this extraneous stuff. Which branch was this PR built off of? See lots of irrelevant changes and conflicts with Beta. Was this built off of the latest Beta?

PR #347, opened 2022-12-29
Asset validation

zerodouglas:
This pull request references #319 and the initial design specification outlined in phonon-network/asset-validation with the following proposed implementation:

```go
valid, err := s.validator.Validate(proposedPhonons)
if err != nil {
  return err
}
```
 
When providing the validator the collection of proposed phonons to validate, each phonon is evaluated based on its currency type. For example, this implementation only supports phonons that are ETH currency (type: 2). This allows any contributor to create a validator for such currency by using the interface defined by "AssetValidator".


chain/ethChainService.go
```diff
@@ -2,138 +2,94 @@ package chain
...
-	redeemable, redeemValue := eth.checkRedeemValue(onChainBalance, suggestedGasPrice)
-	if !redeemable {
-		return "", ErrGasCostExceedsRedeemValue
-	}
+		if balance.Cmp(p.Denomination.Value()) < 0 {
+			log.Error("phonon balance is less than the denomination value")
+			result = append(result, &validator.AssetValidationResult{
+				P:     p,
+				Valid: false,
+				Err:   errors.New("phonon balance is less than what is stated in denomination value"),
```
maevyn11, line 58: should assign this to an error constant.

validator/validator.go
```diff
@@ -2,12 +2,53 @@ package validator
...
+	result := []*AssetValidationResult{}
+
+	if len(proposal) == 0 {
+		return nil, ErrEmptyProposal
+	}
+
+	for _, p := range proposal {
+		switch p.CurrencyType {
+		case 2:
+			result, err := v.ethValidator.Validate(p)
```
maevyn11, line 42: Does this loop work? I think initializing result inside the for loop here will erase the contents of result from the last iteration. Not what we want. Think this is a good case to use named return parameters. 

```suggestion
//In the method sig
func (v *Validator) Validate(proposal []*model.Phonon) (results []*AssetValidationResult, err error) {
...
case 2: 
result, err := v.ethValidator.Validate(p)
...
results = append(results, result)
```

validator/validator.go
```diff
@@ -2,12 +2,53 @@ package validator
...
+	Err   error
+}
+
+type AssetValidator interface {
+	Validate(proposal []*model.Phonon) ([]*AssetValidationResult, error)
+}
+
+type Validator struct {
+	ethValidator   *chain.EthChainService
+	assetValidator AssetValidator
```
maevyn11, line 25: why is the interface a member of this struct?

maevyn11 (changes requested):
Couple small things that need changing, otherwise all the improvements look good.

maevyn11 (changes requested):
Needs to use the sim as the backend by injecting the dependency into EthChainService instead of using Ganache

chain/ethChainService.go
```diff
@@ -7,29 +7,26 @@ import (
...
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
 	ethcrypto "github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/ethclient"
 	log "github.com/sirupsen/logrus"
 )
 
-var ErrGasCostExceedsRedeemValue = errors.New("cannot redeem phonon where gas cost would exceed on chain balance")
-var ErrRedeemAddressInvalid = errors.New("redeem address is invalid")
+var ErrUntrustworthy = errors.New("phonon balance is less than what is stated in denomination value")
```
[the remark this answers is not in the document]
zerodouglas: Agree, I like ErrBalanceTooLow

chain/ethChainService_test.go
```diff
@@ -16,115 +16,200 @@ import (
...
+	}
+
+	highValue, err := model.NewDenomination(big.NewInt(9900000000000))
 	if err != nil {
-		t.Error("unable to fund simEVM. err: ", err)
+		t.Error("unable to add denomination value. err: ", err)
 	}
 
-	fund2, err := fundSimEVM(model.Bitcoin)
+	lowValue, err := model.NewDenomination(big.NewInt(1000000000000000))
```
maevyn11, line 135: maybe add some comments on how many zeroes there are here? It's really hard to tell at a glance whether this value is different than the one on 125

chain/ethChainService_test.go
```diff
@@ -16,115 +16,200 @@ import (
...
+		}
 
-	validator := NewMultiAssetValidator()
-	valid, err := validator.Validate(phonons)
+		if i == 4 {
+			phonon.Denomination = highValue
+		}
+
+		phonons = append(phonons, phonon)
+	}
```
maevyn11, line 177: instead of changing values by index, I suggest it would be a lot easier to read this if you created a bunch of phonons up front with just the different keyIndexes, and then modified the fields directly one at a time before calling validate. As it is, you have to scan up and down from here to the test case check to see what's being done.

chain/ethChainService_test.go
```diff
@@ -16,115 +16,200 @@ import (
...
+
+	valid, err := ethChainSrv.Validate(phonons)
```
maevyn11, line 179: valid should be called validationResult, valid makes it seem like it's a boolean.

chain/ethChainService_test.go
```diff
@@ -16,115 +16,200 @@ import (
...
+		// if r.P.KeyIndex == 2 {
+		// 	t.Logf("validating an invalid phonon with an unsupported chain id: %d", r.P.ChainID)
+		// 	assert.Equal(t, r.Valid, false)
+		// 	assert.EqualError(t, r.Err, model.ErrBalanceTooLow.Error())
+		// }
+
+		if r.P.KeyIndex == 3 {
+			t.Logf("validating an invalid phonon with an unsupported currency type: %s", r.P.CurrencyType)
+			assert.Equal(t, r.Valid, false)
+			assert.EqualError(t, r.Err, model.ErrUnsupportedCurrency.Error())
```
maevyn11, line 206: Would prefer this stuff be checked using regular golang equality statements instead of introducing a new assert dependency, but I know that's more of a style preference, so I'll leave it up to you. 

Suggest like: 
```
if !(r.Valid == false && r.Err == model.ErrUnsupportedCurrency) {
  t.Error("err message")
}
```

model/assetValidator.go
```diff
@@ -6,6 +6,10 @@ import (
 
 var ErrUnsupportedCurrency = errors.New("unsupported currency type")
 var ErrEmptyProposal = errors.New("there were no phonon(s) to validate")
+var ErrBalanceTooLow = errors.New("phonon balance is less than what is stated in denomination value")
+var ErrBalanceTooHigh = errors.New("phonon balance is greater than what is stated in denomination value")
+var ErrInvalidChainID = errors.New("eth chainID unsupported")
+var ErrInvalidEthAddress = errors.New("invalid ethereum address")
```
maevyn11, line 12: this looks nice

orchestrator/session.go
```diff
@@ -88,10 +88,10 @@ var ErrMiningReportNotAvailable = errors.New("could not find mining status repor
...
 // The next step is to run VerifyPIN to gain access to the secure commands on the card
 func NewSession(storage model.PhononCard) (s *Session, err error) {
-	chainSrv, err := chain.NewMultiChainRouter()
-	if err != nil {
-		return nil, err
-	}
+	// chainSrv, err := chain.NewMultiChainRouter()
+	// if err != nil {
+	// 	return nil, err
+	// }
```
maevyn11, line 94: Dead code or does this still have to be integrated into session?

maevyn11 (changes requested):
Mostly looks good but there are a couple concerning elements I commented on.

chain/ethChainService_test.go
```diff
@@ -16,115 +16,200 @@ import (
...
-var simEVM *backends.SimulatedBackend
-var genesisKey *ecdsa.PrivateKey
-var genesisAcct *bind.TransactOpts
-
-func getSimEVM() (*backends.SimulatedBackend, error) {
-	var err error
+func getSimEVM() (simEVM *backends.SimulatedBackend, genesisKey *ecdsa.PrivateKey, genesisAcct *bind.TransactOpts, err error) {
 	if simEVM != nil {
-		return simEVM, nil
+		return simEVM, nil, nil, nil
```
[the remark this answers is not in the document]
zerodouglas: Agreed, it's initialized within the function so it will always be nil. I'll remove it :)

chain/ethChainService_test.go
zerodouglas: I like your suggestion. It removes the dependency of assert and makes it more clear in the end.

orchestrator/session.go
zerodouglas: I can certainly remove it, I commented it out at that time when I removed the package that this function depended on, but simply forgot about it.

chain/ethChainService_test.go
```diff
@@ -16,115 +16,200 @@ import (
...
-	eth.clChainID = testChainID
+	ethChainSrv.cl = sim
 
-	ctx := context.Background()
+	return ethChainSrv, err
+}
 
-	nonce, _, _, err := eth.fetchPreTransactionInfo(ctx, genesisAcct.From)
+func createEthPhonon(desc *model.Phonon) (*model.Phonon, error) {
+	pubKey, err := generatePubKey()
```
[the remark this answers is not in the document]
zerodouglas: Understandable, I had migrated this out since I figured "FundEthPhonon" was not the right name for both creating a phonon and adding funds to it. But now that you've pointed it out, I agree, it's pointless and adds clutter, since the fund function takes the pointer that can just be referenced in the test case.

chain/ethChainService_test.go
zerodouglas: So, by that I suppose you mean just create a bunch of local variables with different values and then value them at a time. Can Do :)

maevyn11: something like that yea

orchestrator/session.go
maevyn11: yea let's remove it, better not to leave clutter in the codebase

chain/ethChainService_test.go
```diff
@@ -122,94 +104,97 @@ func TestValidate(t *testing.T) {
...
 	lowValue, err := model.NewDenomination(big.NewInt(1000000000000000))
 	if err != nil {
 		t.Error("unable to add denomination value. err: ", err)
 	}
 
-	phonons := []*model.Phonon{}
-	for i := 0; i < 10; i++ {
-		phonon, err := createEthPhonon(&model.Phonon{
-			KeyIndex:     model.PhononKeyIndex(i),
+	pubKey, err := generatePubKey()
```
maevyn11, line 125: pubkey needs to loop so that all funds aren't sent to the same address

maevyn11 (changes requested):

chain/ethChainService_test.go
zerodouglas: Fixed in https://github.com/GridPlus/phonon-client/blob/97e6af9466611211b27c5f3e874aeb2f36860fd8/chain/ethChainService_test.go#L177

chain/ethChainService.go
```diff
@@ -35,63 +35,66 @@ func NewEthChainService() (*EthChainService, error) {
...
 
-	for _, p := range proposal {
-		var resultForPhonon *model.AssetValidationResult
+	err := eth.dialRPCNode(uint32(proposal.ChainID))
+	if err != nil {
+		log.Error("could not dial rpc node: ", err)
+		return &model.AssetValidationResult{
+			P:   proposal,
+			Err: err,
+		}, nil
```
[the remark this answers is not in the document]
zerodouglas: Resolution starts https://github.com/GridPlus/phonon-client/blob/97e6af9466611211b27c5f3e874aeb2f36860fd8/chain/ethChainService.go#L49

chain/ethChainService.go
```diff
@@ -47,30 +47,30 @@ func (eth *EthChainService) ValidateSingle(proposal *model.Phonon) (*model.Asset
 	if err != nil {
 		log.Error("could not dial rpc node: ", err)
 		return &model.AssetValidationResult{
-			P:   proposal,
-			Err: err,
-		}, nil
+			P:     proposal,
+			Valid: true,
```
maevyn11, line 51: why is Valid true here? Shouldn't it be false?

chain/ethChainService_test.go
```diff
@@ -104,96 +108,104 @@ func TestValidate(t *testing.T) {
 		t.Error("unable to init simEVM. err: ", err)
 	}
 
-	// 13 zeros
-	value, err := model.NewDenomination(big.NewInt(10000000000000))
+	// 1e zeros
```
maevyn11, line 111: typo?

chain/ethChainService_test.go
```diff
@@ -104,96 +108,104 @@ func TestValidate(t *testing.T) {
...
-			ChainID:      testChainID,
+			p: &model.Phonon{
+				KeyIndex:     3,
+				Denomination: value,
+				ChainID:      testChainID,
+				CurveType:    model.Secp256k1,
+				CurrencyType: model.Ethereum,
+			},
+			valid: false,
+			err:   model.ErrBalanceTooHigh,
```
maevyn11, line 171: nice, this looks much better

chain/ethChainService.go
maevyn11: probably doesn't matter since we're going to check the error. But if in future we don't this could give goofy results.

maevyn11 (approved):
Looks good, a couple small nits.

